PR #41, opened 2018-03-29
Add self-test suite

cdwfs:
To enable:
- Add the [googletest](https://github.com/google/googletest) submodule:
  - `git submodule init`
  - `git submodule update`
- set `SPIRV_REFLECT_BUILD_TESTS` to ON in CMake configuration
- Build & run `test-spirv-reflect project`

Most of the tests focus on verifying the functionality of the public C API (e.g. "if I change a descriptor set number, I'd better be able to retrieve it using its new number"). One short test at the end quietly compares the full YAML output of all test shaders against pre-built "golden YAML", and fails if anything is different. Figuring out exactly what changed is currently a bit of a mess -- basically, re-run build_all_shaders.py to regenerate the golden YAML, and then inspect the git diff to see line-by-line differences.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "third_party/googletest"]
+	path = third_party/googletest
+	url = https://github.com/google/googletest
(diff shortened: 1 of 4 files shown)


cdwfs:
Will resubmit after #44 is merged

PR #56, opened 2018-07-12
Support for multiple SPIRV entrypoints

iburinoc:
This adds support for handling multiple entry points in a single module and walking the static call tree of each to determine which descriptor sets/bindings, interface variables, and push constant blocks are actually used by each.  The interface is a little awkward in places as I didn't want to change behaviour of the single-entry point mode.  Apologies for the large PR, a lot of it is fluff due to essentially creating an extra copy of everything function in the interface.

The tests for the multiple entrypoints are a bit awkward to set up, as glslc and glslang don't support multiple entrypoints, so I set it up as a glsl file with a patch to apply to the disassembled spirv.

Ignore the first commit here, this PR depends on #55, so it should disappear when thats merged.


spirv_reflect.c
```diff
@@ -183,6 +191,96 @@ static uint32_t RoundUp(uint32_t value, uint32_t multiple)
...
+    } else {
+      ++*res_size;
+      ++idx0;
+      ++idx1;
+    }
+  }
+
+  *res = NULL;
+  if (*res_size > 0) {
+    *res = (uint32_t*)calloc(*res_size, sizeof(*res));
```
apazylbe: Shouldn't this be sizeof(**res)?

iburinoc: Fixed, thanks

spirv_reflect.h
```diff
@@ -1026,7 +1419,25 @@ inline const uint32_t* ShaderModule::GetCode() const {
...
+  return m_module.entry_point_count;
+}
+
+/*! @fn GetEntryPointName
+
+  @param index
+  @return
+*/
+inline const char* ShaderModule::GetEntryPointName(uint32_t index) const {
+  return m_module.entry_points[0].name;
```
apazylbe: shouldn't this use index?

iburinoc: Oops, yes.

spirv_reflect.h
```diff
@@ -1057,6 +1468,29 @@ inline SpvReflectResult ShaderModule::EnumerateDescriptorBindings(
   return m_result;
 }
 
+/*! @fn EnumerateEntryPointDescriptorBindings
+
+  @param  entry_point
+  @param  count
+  @param  p_binding_numbers
+  @param  p_set_numbers
```
apazylbe: are these 3 right?

apazylbe: same for the functions below: are the parameter names right?

iburinoc: Good point, the documentation for the original functions I copied from is also wrong, I'll fix those too.

PR #81, opened 2019-11-30
Fixed push constant name

sopyer:
Type name was used before instead of variable name


NikichaTV:
Works greatly, someone has to approve it.

PR #84, opened 2020-02-01
Prevent reading beyond bounds of p_access_chain.

kuwerty:
valgrind threw this out and it does seem like an invalid access.  There's a bounds check on index_index present but it's after the dereference so I'm just guessing at this fix.


chaoticbob:
Hi - can you sign the CLA?

kuwerty:
Sure, it's signed.

PR #103, opened 2020-11-20
Add support for the embedded source code

hliatis:
These changes get the source code from OpSource and OpSourceContinued

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -192,7 +192,7 @@ typedef struct Parser {
   SpvSourceLanguage     source_language;
   uint32_t              source_language_version;
   uint32_t              source_file_id;
-  String                source_embedded;
+  const char*           source_embedded;
   size_t                node_count;
   Node*                 nodes;
   uint32_t              entry_point_count;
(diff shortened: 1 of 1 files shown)


spirv_reflect.c
```diff
@@ -645,6 +647,32 @@ static SpvReflectResult ParseNodes(Parser* p_parser)
         if (p_node->word_count >= 4) {
           CHECKED_READU32(p_parser, p_node->word_offset + 3, p_parser->source_file_id);
         }
+        if (p_node->word_count >= 5) {
+          const char* p_source = (const char*)(p_parser->spirv_code + p_node->word_offset + 4);
+          char* p_source_temp = (char*)calloc(strlen(p_source), sizeof(char*));
```
chaoticbob, line 652: Is it possible for `p_source` to point to an empty string resulting in `strlen` returning 0?

hliatis: I'm not 100% sure. I assumed that if there is a 5th word, then the source code string exists. However, I think I am missing a +1 here.

chaoticbob: IIRC `strcpy` below will copy with the null terminator?

hliatis: +1 added

chaoticbob: Looks good, Thanks.

PR #106, opened 2020-12-08
Fix Werror(s)

jeremy-lunarg:
GCC and Clang were tripping on these werrors.


chaoticbob:
Ooph! Thanks for catching this!

PR #108, opened 2021-01-05
Add printing for non-vector types in common.cpp.

### LINKED ISSUE #68: ParseFormat does not handle doubles correctly
I think ParseFormat should be checking the width of the float/vec types in order to determine the format. https://github.com/chaoticbob/SPIRV-Reflect/blob/8394d3eedd4e62aa1bd7229426256151962539a9/spirv_reflect.c#L2118

Formats like `SPV_REFLECT_FORMAT_R64_SFLOAT`, etc. appear to be missing.

mgsegal:
Also add glsl file that uses doubles.
Test: verified that it works correctly on this file.

Resolves #68


chaoticbob:
I think the ask in #68 is to also have the format of the IO vars parsed as well?

PR #111, opened 2021-01-25
Add missing break statements.

chrismile:
It seems like ParseFormat wrongly identifies 32-bit values as 64-bit values due to two break statements missing in commit https://github.com/KhronosGroup/SPIRV-Reflect/commit/00c4eb534cc5a23b6b941ee8875fda58c34fea3b. This PR should fix the problem.


chaoticbob:
Thank you!

PR #112, opened 2021-02-03
Ensuring embedded source is freed before assignment; e.g. to handle #…

wflohry:
When using an #include extension with glslc, two sources are provided with op = `SpvOpSource` (rather than as `SpvOpSourceContinued`).
Since there is currently only one slot, this new source replaces the existing source_embedded without freeing it.
There might be a better way to handle this (especially for debugging purposes), but at the very least this PR prevents the previous `source_embedded` from leaking.

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -670,6 +670,7 @@ static SpvReflectResult ParseNodes(Parser* p_parser)
           strcpy(p_source_temp, p_source);
       #endif
 
+          SafeFree(p_parser->source_embedded);
           p_parser->source_embedded = p_source_temp;
         }
       }


laurelkeys:
Hey @chaoticbob I've just ran into this same issue (using `glslangValidator` with the `GL_GOOGLE_include_directive` extension) of memory being leaked by the following allocation:
https://github.com/KhronosGroup/SPIRV-Reflect/blob/b68b5a8a5d8ab5fce79e6596f3a731291046393a/spirv_reflect.c#L692

And ensuring that `p_parser->source_embedded` is freed before being assigned `p_source_temp` fixes it. Could this PR be merged?

chaoticbob (approved):

chaoticbob:
Sorry for the super late response. Merging now. Thanks for the PR.

PR #118, opened 2021-05-26
Report result type for OpTypeImage

kevin-mccullough:
This PR adds missing support for reporting the result type for OpTypeImages.


chaoticbob:
Looks good to go, Thanks, Kevin.

PR #119, opened 2021-06-19
Fix array type stride information

nyorain:
By not forwarding struct member decorations, information is currently lost, e.g. the stride of matrix types inside an array will always be 0. There might be more similar cases, I added the forwarding of the struct member decorations to all nested types. And also added a testcase for this.

My editor automatically removes whitespace, I split that off into a separate commit, the real changes are only in the second commit. If you want to keep the whitespace, I'll kick the commit out.

By the way: To run the tests I had to update googletest to master, GCC 11 outputs a warning and fails to compile (Werror) with the referenced version. And even after regenerating all shaders and golden yaml files I still get 2 errors in hlsl tests unrelated to this change.


spencer-lunarg:
@nyorain what is the status of this PR? I realize you just added a test, but not sure what code change was made to `spirv_reflect.c` itself

(also clang-format was added, so trailing whitespace is fixed now)

nyorain:
Uh seems like the fix itself got lost somehow. I think it was forwarding `p_struct_member_decorations` in recursive ParseType calls (e.g. for vector/matrix/sampleImage/(runtime)array/pointer) as in [this](https://github.com/nyorain/vil/commit/c890aeda3ec55d2b251fa26cc9df1f84dacef70d#diff-378a103a2f84d1037bf9dc35c95c216d8d933b023b6f1b640b3b26b063908807) commit.
I'm not using spirv_reflect anymore, might have been fixed in the meantime. Feel free to add the test and fix as needed.

PR #126, opened 2021-09-27
Add accessed pointers through OpImageTexelPointer to entrypoint

MarijnS95:
Without this spvReflectEnumerateEntryPointDescriptorBindings omits all descriptor bindings from a function that are only accessed through `SpvOpImageTexelPointer`.


chaoticbob:
Thanks!

PR #127, opened 2021-09-27
Fix use of strcat_s when parsing SpvOpSourceContinued

gmueckl:
Pass the correct length of the target buffer to strcat_s.


chaoticbob:
Sorry for the late response. Thanks for the PR.

chaoticbob (approved):

PR #129, opened 2021-10-16
Removed null test in SafeFree Macro

blackbird806:
Freeing a null pointer does nothing therefore it is redundant to check null before free.


chaoticbob:
Thanks! Old habits hard to break.

PR #133, opened 2022-01-06
Fix Segfault when parsing runtime array of array of struct

tomix1024:
TL;DR When parsing a block variable that is a runtime array of array of struct, e.g. `Foo var[][42];` a segmentation fault was produced because the `p_type_node` was not correctly resolved in `ParseDescriptorBlockVariable`. This pull request fixes the problem.

When parsing a block variable the node representing the underlying type is looked up in order to parse potential member variables if the base type of the array is a struct.
The `SpvReflectPrvNode` nodes of array types don't have information about potential struct members associated to them, while the type description `SpvReflectTypeDescription` does.
Therefore the node corresponding to the base non-array type is resolved.
For multiple levels of arrays of arrays this was correctly implemented.
However, for runtime arrays, only one level of array-ness (the runtime array) was removed.
In the event of parsing a runtime array of array of struct the empty `p_type_node->member_*` fields of the array type node were accessed.


chaoticbob:
Thanks for the fix!

Issue #135, opened 2022-01-22
Updating main_io_variables.cpp example to take into consideration GLSL built-in variables


Lucodivo:
I'm currently going through the example of pulling input variables from a SPIR-V vertex shader. The original GLSL vertex shader utilizes the built-in variable `gl_InstanceIndex`. This shows up as one of the input variables returned from `spvReflectEnumerateInputVariables()` at `location = 4294967295`. Although this may be correct in terms of what input the shader is receiving, the example will allow it to add 4 bytes to the stride of a VkVertexInputAttributeDescription. Which is incorrect and may be a common bug while using this example. 

I've also noticed that using `--reflect` with the SPIRV-Cross command line tool doesn't consider `gl_InstanceIndex` as an input variable.

For reference, this is the shader in question that I used with `spvReflectCreateShaderModule()` and SPIRV-Cross

```
#version 460

layout (location = 0) in vec3 vPosition;
layout (location = 1) in vec3 vNormal;
layout (location = 2) in vec3 vColor;
layout (location = 3) in vec2 vTexCoord;

layout (location = 0) out vec3 outColor;
layout (location = 1) out vec2 texCoord;

layout(set = 0, binding = 0) uniform CameraBuffer{
	mat4 view;
	mat4 proj;
	mat4 viewproj;
} cameraData;

struct ObjectData {
	mat4 model;
};
layout(std140, set = 1, binding = 0) readonly buffer ObjectBuffer {
	ObjectData objects[];
} objectBuffer;

void main()
{
	mat4 modelMatrix = objectBuffer.objects[gl_InstanceIndex].model;
	mat4 transformMatrix = cameraData.viewproj * modelMatrix;
	gl_Position = transformMatrix * vec4(vPosition, 1.0f);
	outColor = vColor;
	texCoord = vTexCoord;
}
```

https://github.com/KhronosGroup/SPIRV-Reflect/blob/4689b3360cb38283b67926a065a0f2cc285928b5/examples/main_io_variables.cpp#L204

P.S. Thank you for all your work!!

Lucodivo:
For my solution, ignoring the variable if `spvReflectInterfaceVariable.built_in == -1` gives me the expected result.

chaoticbob:
Hi - sorry this gave you a bit of grief. This is an unfortunate side effect exposing the built-ins for tooling. Glad you found a solution. I'll make some updates to the documentation so this is more clear.

PR #137, opened 2022-01-22
ignore built-in vars for vertex input description

Lucodivo:
#135


chaoticbob:
Hi - is the intent here for the spirv-reflect executable to not display input variable as a vertex attribute if it's a built-in?

Lucodivo:
The code I edited is only an example on how to use the SPIR-V Reflect library. Specifically it is trying to fill in the information of a VkVertexInputBindingDescription struct based on the reflection data. This information should not include built-in variables. I noticed the problem in my own code when using this example because the built-in variable was added to the stride of my VkVertexInputBindingDescription struct. Which is incorrect.

chaoticbob:
I totally missed that. Thanks! This makes sense.

PR #149, opened 2022-06-13
Update googletest submodule to release-1.11.0

vettoreldaniele:
Update googletest to a recent release (`release-1.11.0`), since tests do not build with a recent clang version due to googletest being too old.


chaoticbob:
Thanks!

Issue #121, opened 2021-07-09
Boolean Specialization Constants


reduz:
I am checking SPIRV-Reflect, but I see nowhere where the list of boolean specialization constants can be obtained from a stage. Is this supported?

reduz:
Ended up implementing this in our local fork, for anyone interested:
https://github.com/godotengine/godot/pull/50325

If there is any interest in me submitting a PR with my changes, please let me know.

chaoticbob:
Hi,

Specialization constants haven't gotten a lot of love since unfortunately. It's on my list of thing sto investigate but haven't gotten to it yet. It would be greatly appreciated if you could submit a PR. It would at least get teh ball rolling on specialization constants.

Thanks,
Hai

reduz:
Alright, a PR will be submitted soon!

chaoticbob:
Thanks! Looking forward to it.

pushrax:
I'm about to implement this in another engine, @reduz any chance of a PR any time soon?

PR #154, opened 2022-07-28
Adding preliminary `SpecConstant` support [WIP]

shangjiaxuan:
Used code form godot/godot as starting point. See https://github.com/KhronosGroup/SPIRV-Reflect/issues/121

Next steps: 64 bit types need 2 words as operand, `SpecConstantOp` need implementation for evaluating spec constant (spirv-cross has an implementation for uint https://github.com/KhronosGroup/SPIRV-Cross/pull/1463). Also maybe composite type evaluation support?


shangjiaxuan:
CLA may need contributers to `https://github.com/godotengine/godot/pull/50325`. (Godot is licensed under MIT license, so maybe the agreement is compatible? Not a law expert...)

shangjiaxuan:
Related issue #121 (From godot, where starting of current pull is based on)

shangjiaxuan:
There may be better ways to evaluate the values (e.g. using some jit compiler that compiles to cpu native code), not sure if all the switch is easily maintainable (and if if it's possible through some `#define`s). Will add test cases for maintenance.

shangjiaxuan:
Currently using the following test from spirv-cross will result in correct specialized array size.
```
#version 460

layout(constant_id = 6) const uint UTWO = 2;

const uint IMUL = UTWO * UTWO;               // 4

layout(std430, set = 0, binding = 2)buffer SSBO_IMul{ float val[IMUL - 3]; float dummy; } IMul;

layout(local_size_x_id = 1, local_size_y_id = 2, local_size_z_id = 3) in;
void main(void)
{}
```
spirv-reflect output:
```
Descriptor bindings: 1

    Binding 0.2
      spirv id : 15
      set      : 0
      binding  : 2
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : false
      name     : IMul (SSBO_IMul)
          // size = 0, padded size = 0
          struct SSBO_IMul {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4 UNUSED
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } IMul;
```
Changing constant value should work, but not tested.
Maybe storing a specconstantop dictionary can make most duplicate result evaluation go away. That may be included (compute spec ops after user specify a current value , or lazy init?)?

Still WIP.

shangjiaxuan:
Moved much of recurring code into `#define`s, also scalar integer arithmetic should be working now. Boolean and vector types still unimplemented.

shangjiaxuan:
Scalar constants should work now. Tested with following glsl from spirv-cross:
```
#version 450

layout(local_size_x = 1) in;

layout(constant_id = 0) const bool TRUE = true;
layout(constant_id = 1) const bool FALSE = false;
layout(constant_id = 2) const int SONE = 1;
layout(constant_id = 3) const int STWO = 2;
layout(constant_id = 4) const int SNEG_TWO = -2;
layout(constant_id = 5) const uint UONE = 1;
layout(constant_id = 6) const uint UTWO = 2;
layout(constant_id = 7) const int SNEG_THREE = -3;

const uint IADD = SONE + STWO + UONE + UTWO; // 6
const uint ISUB = UTWO - SONE;               // 1
const uint IMUL = UTWO * UTWO;               // 4
const uint UDIV = UTWO / UTWO;               // 1
const int SDIV = STWO / SNEG_TWO;            // -1
//const int SREM = STWO % SNEG_THREE;          // 1
const int SREM = 1;
const int SMOD = STWO % SNEG_THREE;          // -1
const uint UMOD = IADD % IMUL;               // 2

const uint LSHL = IADD << ISUB;              // 12
const uint RSHL = IADD >> ISUB;              // 3
const int RSHA = (-int(IADD)) >> (-SDIV);         // -3

const bool IEQ = IADD == ISUB;               // false
const bool INEQ = IADD != ISUB;              // true
const bool ULT = IADD < ISUB;                // false
const bool ULE = IADD <= ISUB;               // false
const bool UGT = IADD > ISUB;                // true
const bool UGE = IADD >= ISUB;               // true

const bool SLT = SMOD < SREM;                // true
const bool SLE = SMOD <= SREM;               // true
const bool SGT = SMOD > SREM;                // false
const bool SGE = SMOD >= SREM;               // false

const bool LOR = IEQ || SLT;                 // true
const bool LAND = IEQ && SLT;                // false
const bool LNOT = !LOR;                      // false

const uint AND = IADD & IADD;                // 6
const uint OR = IADD | ISUB;                 // 7
const uint XOR = IADD ^ IADD;                // 0
const uint NOT = ~XOR;                       // UINT_MAX

const bool LEQ = LAND == LNOT;               // true
const bool LNEQ = LAND != LNOT;              // false

const uint SEL = IEQ ? IADD : ISUB;          // 1

#define DUMMY_SSBO(name, bind, size) layout(std430, set = 0, binding = bind) buffer SSBO_##name { float val[size]; float dummy; } name

// Normalize all sizes to 1 element so that the default offsets in glslang matches up with what we should be computing.
// If we do it right, we should get no layout(offset = N) expressions.
DUMMY_SSBO(IAdd, 0, IADD - 5);
DUMMY_SSBO(ISub, 1, ISUB);
DUMMY_SSBO(IMul, 2, IMUL - 3);
DUMMY_SSBO(UDiv, 3, UDIV);
DUMMY_SSBO(SDiv, 4, SDIV + 2);
DUMMY_SSBO(SRem, 5, SREM);
DUMMY_SSBO(SMod, 6, SMOD + 2);
DUMMY_SSBO(UMod, 7, UMOD - 1);
DUMMY_SSBO(LShl, 8, LSHL - 11);
DUMMY_SSBO(RShl, 9, RSHL - 2);
DUMMY_SSBO(RSha, 10, RSHA + 4);
DUMMY_SSBO(IEq, 11, IEQ ? 2 : 1);
DUMMY_SSBO(INeq, 12, INEQ ? 1 : 2);
DUMMY_SSBO(Ult, 13, ULT ? 2 : 1);
DUMMY_SSBO(Ule, 14, ULE ? 2 : 1);
DUMMY_SSBO(Ugt, 15, UGT ? 1 : 2);
DUMMY_SSBO(Uge, 16, UGE ? 1 : 2);
DUMMY_SSBO(Slt, 17, SLT ? 1 : 2);
DUMMY_SSBO(Sle, 18, SLE ? 1 : 2);
DUMMY_SSBO(Sgt, 19, SGT ? 2 : 1);
DUMMY_SSBO(Sge, 20, SGE ? 2 : 1);
DUMMY_SSBO(Lor, 21, LOR ? 1 : 2);
DUMMY_SSBO(Land, 22, LAND ? 2 : 1);
DUMMY_SSBO(Lnot, 23, LNOT ? 2 : 1);
DUMMY_SSBO(And, 24, AND - 5);
DUMMY_SSBO(Or, 24, OR - 6);
DUMMY_SSBO(Xor, 24, XOR + 1);
DUMMY_SSBO(Not, 25, NOT - 0xfffffffeu);
DUMMY_SSBO(Leq, 26, LEQ ? 1 : 2);
DUMMY_SSBO(Lneq, 27, LNEQ ? 2 : 1);
DUMMY_SSBO(Sel, 28, SEL);

void main()
{
	IAdd.val[0] = 0.0;
	ISub.val[0] = 0.0;
	IMul.val[0] = 0.0;
	UDiv.val[0] = 0.0;
	SDiv.val[0] = 0.0;
	SRem.val[0] = 0.0;
	SMod.val[0] = 0.0;
	UMod.val[0] = 0.0;
	LShl.val[0] = 0.0;
	RShl.val[0] = 0.0;
	RSha.val[0] = 0.0;
	IEq.val[0] = 0.0;
	INeq.val[0] = 0.0;
	Ult.val[0] = 0.0;
	Ule.val[0] = 0.0;
	Ugt.val[0] = 0.0;
	Uge.val[0] = 0.0;
	Slt.val[0] = 0.0;
	Sle.val[0] = 0.0;
	Sgt.val[0] = 0.0;
	Sge.val[0] = 0.0;
	Lor.val[0] = 0.0;
	Land.val[0] = 0.0;
	Lnot.val[0] = 0.0;
	And.val[0] = 0.0;
	Or.val[0] = 0.0;
	Xor.val[0] = 0.0;
	Not.val[0] = 0.0;
	Leq.val[0] = 0.0;
	Lneq.val[0] = 0.0;
	Sel.val[0] = 0.0;
}
```
Reflection output:
```
generator       : Google Shaderc over Glslang
source lang     : GLSL
source lang ver : 450
source file     :
entry point     : main (stage=CS)
local size      : (1, 1, 1)


  Sepecialization constants: 8

    0:
      spirv id   : 8
      constant id: 2
      name       : SONE
      type       : signed 32 bit integer
      default    : 1
    1:
      spirv id   : 9
      constant id: 3
      name       : STWO
      type       : signed 32 bit integer
      default    : 2
    2:
      spirv id   : 14
      constant id: 5
      name       : UONE
      type       : unsigned 32 bit integer
      default    : 1
    3:
      spirv id   : 16
      constant id: 6
      name       : UTWO
      type       : unsigned 32 bit integer
      default    : 2
    4:
      spirv id   : 49
      constant id: 4
      name       : SNEG_TWO
      type       : signed 32 bit integer
      default    : -2
    5:
      spirv id   : 64
      constant id: 7
      name       : SNEG_THREE
      type       : signed 32 bit integer
      default    : -3
    6:
      spirv id   : 251
      constant id: 0
      name       : TRUE
      type       : boolean
      default    : 1
    7:
      spirv id   : 252
      constant id: 1
      name       : FALSE
      type       : boolean
      default    : 0


  Descriptor bindings: 31

    Binding 0.0
      spirv id : 23
      set      : 0
      binding  : 0
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : IAdd (SSBO_IAdd)
          // size = 0, padded size = 0
          struct SSBO_IAdd {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } IAdd;


    Binding 0.1
      spirv id : 33
      set      : 0
      binding  : 1
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : ISub (SSBO_ISub)
          // size = 0, padded size = 0
          struct SSBO_ISub {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } ISub;


    Binding 0.2
      spirv id : 41
      set      : 0
      binding  : 2
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : IMul (SSBO_IMul)
          // size = 0, padded size = 0
          struct SSBO_IMul {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } IMul;


    Binding 0.3
      spirv id : 47
      set      : 0
      binding  : 3
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : UDiv (SSBO_UDiv)
          // size = 0, padded size = 0
          struct SSBO_UDiv {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } UDiv;


    Binding 0.4
      spirv id : 56
      set      : 0
      binding  : 4
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : SDiv (SSBO_SDiv)
          // size = 0, padded size = 0
          struct SSBO_SDiv {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } SDiv;


    Binding 0.5
      spirv id : 62
      set      : 0
      binding  : 5
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : SRem (SSBO_SRem)
          // size = 0, padded size = 0
          struct SSBO_SRem {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } SRem;


    Binding 0.6
      spirv id : 70
      set      : 0
      binding  : 6
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : SMod (SSBO_SMod)
          // size = 0, padded size = 0
          struct SSBO_SMod {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } SMod;


    Binding 0.7
      spirv id : 77
      set      : 0
      binding  : 7
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : UMod (SSBO_UMod)
          // size = 0, padded size = 0
          struct SSBO_UMod {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } UMod;


    Binding 0.8
      spirv id : 85
      set      : 0
      binding  : 8
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : LShl (SSBO_LShl)
          // size = 0, padded size = 0
          struct SSBO_LShl {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } LShl;


    Binding 0.9
      spirv id : 93
      set      : 0
      binding  : 9
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : RShl (SSBO_RShl)
          // size = 0, padded size = 0
          struct SSBO_RShl {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } RShl;


    Binding 0.10
      spirv id : 104
      set      : 0
      binding  : 10
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : RSha (SSBO_RSha)
          // size = 0, padded size = 0
          struct SSBO_RSha {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } RSha;


    Binding 0.11
      spirv id : 113
      set      : 0
      binding  : 11
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : IEq (SSBO_IEq)
          // size = 0, padded size = 0
          struct SSBO_IEq {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } IEq;


    Binding 0.12
      spirv id : 120
      set      : 0
      binding  : 12
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : INeq (SSBO_INeq)
          // size = 0, padded size = 0
          struct SSBO_INeq {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } INeq;


    Binding 0.13
      spirv id : 127
      set      : 0
      binding  : 13
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Ult (SSBO_Ult)
          // size = 0, padded size = 0
          struct SSBO_Ult {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Ult;


    Binding 0.14
      spirv id : 134
      set      : 0
      binding  : 14
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Ule (SSBO_Ule)
          // size = 0, padded size = 0
          struct SSBO_Ule {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Ule;


    Binding 0.15
      spirv id : 141
      set      : 0
      binding  : 15
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Ugt (SSBO_Ugt)
          // size = 0, padded size = 0
          struct SSBO_Ugt {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Ugt;


    Binding 0.16
      spirv id : 148
      set      : 0
      binding  : 16
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Uge (SSBO_Uge)
          // size = 0, padded size = 0
          struct SSBO_Uge {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Uge;


    Binding 0.17
      spirv id : 155
      set      : 0
      binding  : 17
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Slt (SSBO_Slt)
          // size = 0, padded size = 0
          struct SSBO_Slt {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Slt;


    Binding 0.18
      spirv id : 162
      set      : 0
      binding  : 18
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Sle (SSBO_Sle)
          // size = 0, padded size = 0
          struct SSBO_Sle {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Sle;


    Binding 0.19
      spirv id : 169
      set      : 0
      binding  : 19
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Sgt (SSBO_Sgt)
          // size = 0, padded size = 0
          struct SSBO_Sgt {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Sgt;


    Binding 0.20
      spirv id : 176
      set      : 0
      binding  : 20
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Sge (SSBO_Sge)
          // size = 0, padded size = 0
          struct SSBO_Sge {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Sge;


    Binding 0.21
      spirv id : 183
      set      : 0
      binding  : 21
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Lor (SSBO_Lor)
          // size = 0, padded size = 0
          struct SSBO_Lor {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Lor;


    Binding 0.22
      spirv id : 190
      set      : 0
      binding  : 22
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Land (SSBO_Land)
          // size = 0, padded size = 0
          struct SSBO_Land {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Land;


    Binding 0.23
      spirv id : 197
      set      : 0
      binding  : 23
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Lnot (SSBO_Lnot)
          // size = 0, padded size = 0
          struct SSBO_Lnot {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Lnot;


    Binding 0.24
      spirv id : 204
      set      : 0
      binding  : 24
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : And (SSBO_And)
          // size = 0, padded size = 0
          struct SSBO_And {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } And;


    Binding 0.24
      spirv id : 212
      set      : 0
      binding  : 24
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Or (SSBO_Or)
          // size = 0, padded size = 0
          struct SSBO_Or {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Or;


    Binding 0.24
      spirv id : 219
      set      : 0
      binding  : 24
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Xor (SSBO_Xor)
          // size = 0, padded size = 0
          struct SSBO_Xor {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Xor;


    Binding 0.25
      spirv id : 227
      set      : 0
      binding  : 25
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Not (SSBO_Not)
          // size = 0, padded size = 0
          struct SSBO_Not {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Not;


    Binding 0.26
      spirv id : 234
      set      : 0
      binding  : 26
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Leq (SSBO_Leq)
          // size = 0, padded size = 0
          struct SSBO_Leq {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Leq;


    Binding 0.27
      spirv id : 241
      set      : 0
      binding  : 27
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Lneq (SSBO_Lneq)
          // size = 0, padded size = 0
          struct SSBO_Lneq {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Lneq;


    Binding 0.28
      spirv id : 247
      set      : 0
      binding  : 28
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : true
      name     : Sel (SSBO_Sel)
          // size = 0, padded size = 0
          struct SSBO_Sel {
              float val[1]; // abs offset = 0, rel offset = 0, size = 4, padded size = 4, array stride = 4
              float dummy;  // abs offset = 4, rel offset = 4, size = 4, padded size = 4 UNUSED
          } Sel;
```

shangjiaxuan:
Next steps would be cleaning up clutter, removing dependency on parser and better evaluation. 
Copy needed node info into a internal array, each result id have a evaluation state flag, current value, type pointe, and list of related spec constants.

shangjiaxuan:
Implementation is now done, but the API needs clearing up.

shangjiaxuan:
Implementation done, adding test cases...

spirv_reflect.h
```diff
@@ -79,6 +80,9 @@ typedef enum SpvReflectResult {
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_BLOCK_MEMBER_REFERENCE,
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_ENTRY_POINT,
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_EXECUTION_MODE,
+  SPV_REFLECT_RESULT_ERROR_SPIRV_DUPLICATE_SPEC_CONSTANT_NAME,
```
shangjiaxuan, line 83: These error codes are somewhat inconsistent in return value, maybe using a single evaluation error is better.

spirv_reflect.h
```diff
@@ -92,10 +96,20 @@ SPV_REFLECT_MODULE_FLAG_NO_COPY - Disables copying of SPIR-V code
   This is flag is intended for cases where the memory overhead of
   storing the copied SPIR-V is undesirable.
 
+SPV_REFLECT_MODULE_FLAG_EVALUATE_CONSTANT - copies constant 
+  instructions for use with later evaluation.
+
+SPV_REFLECT_MODULE_FLAG_EVALUATE_NO_COPY - use p_code for later
```
shangjiaxuan, line 102: Need to remove these unneeded comment

spirv_reflect.h
```diff
@@ -257,6 +271,7 @@ typedef enum SpvReflectGenerator {
 } SpvReflectGenerator;
 
 enum {
+  SPV_REFLECT_MAX_VECTOR_DIMS                   =  4,
```
shangjiaxuan, line 274: Vector16 capability allows 8 and 16 components.

spirv_reflect.h
```diff
@@ -451,11 +525,25 @@ typedef struct SpvReflectEntryPoint {
     uint32_t                        x;
     uint32_t                        y;
     uint32_t                        z;
+    int                             flags; // 0 if just fixed local size
```
shangjiaxuan, line 528: This addition will change offset of later members and structure size. Will be an abi change.

spirv_reflect.h
```diff
@@ -1422,6 +1544,94 @@ const char* spvReflectBlockVariableTypeName(
...
+/*! @fn spvReflectGetSpecConstantValue
+@brief  Get the current value of specialization constant. Type must follow c/c++ type aliasing rules.
+@param  p_eval      Pointer to an instance of SpvReflectEvaluation.
+@param  specId      Specialization constant id of the constant.
+@param  value       Current value.
+@return             If successful, returns SPV_REFLECT_RESULT_SUCCESS.
+Otherwise, the error code indicates the cause of
+the failure.
+*/
+SpvReflectResult spvReflectGetSpecConstantValue(SpvReflectEvaluation* p_eval, uint32_t specId, const SpvReflectValue** value);
```
shangjiaxuan, line 1605: Get and set have different parameter types. Unifying may be better?

tests/spec_constant/rebuild.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Creates multi_entrypoint.spv from multi_entrypoint.glsl and
```
shangjiaxuan, line 2: Comment need updating

tests/spec_constant/test_32bit.spv.dis.patch
```diff
@@ -0,0 +1,14 @@
...
+<                OpExecutionMode %main LocalSize 1 1 1
+---
+>                ; should ignore this if working with WorkGroupSize builtin
+>                OpExecutionMode %main LocalSize 2 3 4
+425c426
+<        %SREM = OpSpecConstantOp %int SMod %STWO %SNEG_THREE
+---
+>        %SREM = OpSpecConstantOp %int SRem %STWO %SNEG_THREE
+601c602,603
+<         %262 = OpSpecConstant %uint 1
```
shangjiaxuan, line 11: These numbers will change after changing glsl code. Patch will need to be updated

spirv_reflect.c
```diff
@@ -227,7 +299,7 @@ static uint32_t RoundUp(
    uint32_t value,
    uint32_t multiple)
 {
-  assert(multiple && ((multiple & (multiple - 1)) == 0));
+  SPV_REFLECT_ASSERT(multiple && ((multiple & (multiple - 1)) == 0));
```
shangjiaxuan, line 302: These assert changes are for respecting the cmake option.

spirv_reflect.c
```diff
@@ -363,13 +435,13 @@ static bool InRange(
 }
 
 static SpvReflectResult ReadU32(
-  SpvReflectPrvParser* p_parser,
+  const SpvReflectPrvParser* p_parser,
```
shangjiaxuan, line 438: Parser needs to be constant from needed call tree. Constant doesn't change any behavior.

spirv_reflect.c
```diff
@@ -521,7 +593,37 @@ static SpvReflectPrvNode* FindNode(
...
+    SpvReflectPrvEvaluationNode* p_elem = &(p_eval->nodes[i]);
+    if (p_elem->specId == specid) {
+      p_node = p_elem;
+      break;
+    }
+  }
+  return p_node;
+}
+
+static SpvReflectTypeDescription* FindType(const SpvReflectShaderModule* p_module, uint32_t type_id)
```
shangjiaxuan, line 626: Module can also be const. Since this may be called in evaluation (when module is initialized and const), adding const here is needed.

spirv_reflect.c
```diff
@@ -824,12 +927,7 @@ static SpvReflectResult ParseNodes(SpvReflectPrvParser* p_parser)
       case SpvOpConstant:
       case SpvOpConstantComposite:
       case SpvOpConstantSampler:
-      case SpvOpConstantNull: {
-        CHECKED_READU32(p_parser, p_node->word_offset + 1, p_node->result_type_id);
-        CHECKED_READU32(p_parser, p_node->word_offset + 2, p_node->result_id);
-      }
-      break;
-
+      case SpvOpConstantNull:
```
shangjiaxuan, line 930: These can fall through. Putting `SpvOpConstant` and `SpvOpSpecConstant` seems more logical

spirv_reflect.c
```diff
@@ -1214,6 +1314,9 @@ static SpvReflectResult ParseFunctions(SpvReflectPrvParser* p_parser)
       }
       p_func->callee_ptrs = (SpvReflectPrvFunction**)calloc(p_func->callee_count,
                                                             sizeof(*(p_func->callee_ptrs)));
+      if(IsNull(p_func->callee_ptrs)){
```
shangjiaxuan, line 1317: Added a few allocation failure checks. But still, accessing internal members before testing for NULL before SafeFree parents still may result in segfault on creation alloc failure.

spirv_reflect.c
```diff
@@ -1314,7 +1417,8 @@ static SpvReflectResult ParseNames(SpvReflectPrvParser* p_parser)
   return SPV_REFLECT_RESULT_SUCCESS;
 }
 
-static SpvReflectResult ParseDecorations(SpvReflectPrvParser* p_parser)
+// builtin parsing should be done with shader module (if global) or entry point (if local)
+static SpvReflectResult ParseDecorations(SpvReflectShaderModule* p_module, SpvReflectPrvParser* p_parser)
```
shangjiaxuan, line 1421: Add parameter for storing needed builtin info. One result_id is not guaranteed to have only one builtin decoration.

spirv_reflect.c
```diff
@@ -1437,7 +1542,18 @@ static SpvReflectResult ParseDecorations(SpvReflectPrvParser* p_parser)
       case SpvDecorationBuiltIn: {
         p_target_decorations->is_built_in = true;
         uint32_t word_offset = p_node->word_offset + member_offset + 3;
-        CHECKED_READU32_CAST(p_parser, word_offset, SpvBuiltIn, p_target_decorations->built_in);
+        // no rule specifies a result cannot be decorated twice. But let's assume this for now...
```
shangjiaxuan, line 1545: Comment not updated...

spirv_reflect.c
```diff
@@ -3332,6 +3494,282 @@ static SpvReflectResult ParseExecutionModes(
...
+    (((op_code) == SpvOpConstantTrue) || ((op_code) == SpvOpConstantFalse) || ((op_code) == SpvOpConstant))
+#define IS_CONSTANT_LITERAL_OP(op_code) (IS_SPEC_LITERAL_OP(op_code)||IS_LITERAL_OP(op_code))
+// constant null and constant sampler etc are not handled in literals or anywhere for now...
+#define IS_CONSTANT_OP(op_code) \
+    (IS_CONSTANT_LITERAL_OP(op_code) || ((op_code) == SpvOpSpecConstantOp) || ((op_code) == SpvOpSpecConstantComposite) \
+    || ((op_code) == SpvOpConstantComposite) || ((op_code) == SpvOpConstantNull) || ((op_code) == SpvOpConstantSampler) \
+    || ((op_code) == SpvOpConstantPipeStorage) || ((op_code) == SpvOpSpecConstantCompositeContinuedINTEL)\
+    || ((op_code) == SpvOpConstantFunctionPointerINTEL) || ((op_code) == SpvOpConstantCompositeContinuedINTEL))
+
+#if SPIRV_REFLECT_ENABLE_CONSTANT_EVALUATION
```
shangjiaxuan, line 3551: Define SPIRV_REFLECT_ENABLE_CONSTANT_EVALUATION to 1 to enable evaluation functions to compile. Runtime enable is currently done through flag.

spirv_reflect.c
```diff
@@ -3332,6 +3494,282 @@ static SpvReflectResult ParseExecutionModes(
...
+#ifdef _MSC_VER
+#define SPIRV_REFLECT_FORCEINLINE __forceinline
```
shangjiaxuan, line 3553: FORCEINLINE can make the evaluation kernel function compile into multiple function bodies that have no overhead of inner switches. Advantage over writing multiple functions is that state tracking as here in two preprocessing passes and in evaluation is done in the same place. This can save a lot of going up and down through code. Easier to maintain.

spirv_reflect.c
```diff
@@ -3332,6 +3494,282 @@ static SpvReflectResult ParseExecutionModes(
...
+
+#endif
+
+static SpvReflectResult ParseSpecializationConstants(SpvReflectPrvParser* p_parser, SpvReflectShaderModule* p_module)
+{
+  p_module->specialization_constant_count = 0;
+  p_module->specialization_constants = NULL;
+
+  uint32_t constant_instruction_num = 0;
+  uint32_t instruction_size = 0;
```
shangjiaxuan, line 3586: No longer needed

spirv_reflect.c
```diff
@@ -3896,6 +4343,16 @@ void spvReflectDestroyShaderModule(SpvReflectShaderModule* p_module)
   if ((p_module->_internal->module_flags & SPV_REFLECT_MODULE_FLAG_NO_COPY) == 0) {
     SafeFree(p_module->_internal->spirv_code);
   }
+
+  SafeFree(p_module->specialization_constants);
+
+  if (p_module->_internal->module_flags & SPV_REFLECT_MODULE_FLAG_EVALUATE_CONSTANT) {
```
shangjiaxuan, line 4349: If internal failed to allocate, this will segfault.

common/output_stream.cpp
```diff
@@ -599,7 +599,9 @@ std::string ToStringComponentType(const SpvReflectTypeDescription& type, uint32_
   return ss.str();
 }
 
-void ParseBlockMembersToTextLines(const char* indent, int indent_depth, bool flatten_cbuffers, const std::string& parent_name, uint32_t member_count, const SpvReflectBlockVariable* p_members, std::vector<TextLine>* p_text_lines)
+void ParseBlockMembersToTextLines(
```
shangjiaxuan, line 602: Struct member may have specop array size. Add non-const eval parameter.

spirv_reflect.h
```diff
@@ -1450,6 +1663,7 @@ class ShaderModule {
   SpvReflectResult GetResult() const;
 
   const SpvReflectShaderModule& GetShaderModule() const;
+  Evaluation GetEvaluation() const;
```
shangjiaxuan, line 1666: Constant module can retrieve non-const evaluation interface. Evaluation state should be mutable even after module is fully parsed. (on-demand evaluation, set constant, re-evaluate)

shangjiaxuan (commented):
Should be clean now

shangjiaxuan:
Bazel build does not have all the preprocessor definition options (both this and the assert definition), yaml have no specialization op support. But I suppose yaml doesn't have a way to store constant expression ast either.

chaoticbob:
Hi, thanks for this massive change. However, I need to check to make sure that it's possible to accept MIT license originally from godot.

On a different note, SPIRV-Reflect has a pretty set coding style. Could you make sure the alignmet of structs and code follow the existing convention?

Thanks,
Hai

shangjiaxuan:
@chaoticbob I asked the original author of the original spec-constant part, and confirmed the license https://github.com/godotengine/godot/pull/50325. Just found some inconsistencies in indentations I didn't find when preparing to open this pr, edited that.

chaoticbob:
@shangjiaxuan Thanks for checking with on the godot licensing. Very much appreciated.

spirv_reflect.h
```diff
@@ -327,9 +339,68 @@ typedef struct SpvReflectTypeDescription {
...
   struct SpvReflectTypeDescription* members;
+
+  // for vector, array and matrix types
+  uint32_t                          component_type_id;
 } SpvReflectTypeDescription;
 
 
+typedef struct SpvReflectScalarValueData {
+  union {
+    /* small types not implemented yet... */
```
chaoticbob, line 350: Indentiation for member names should be aligned.

chaoticbob (commented):
Just a comment on indentation for a struct.

shangjiaxuan:
Since this library mainly interfaces with vulkan, and vulkan passes specialization info through `VkSpecializationMapEntry` and `VkSpecializationInfo`, it should be useful to have an api that generates this from internal data (all evaluation node and value are from the same allocation already).
```
/*
*    @brief retrieves all specialization constants that have been modified ready to be passed to vulkan api
*        call with pMapEntries == null to retrieve number of entries, and then call with allocated space to retrieve
*        the actual memory mappings. Function will internally const-cast the VkSpecializationMapEntry array on second
*        call.
*    @param info : the info to be passed into vulkan api
*    @return        SPIRV_REFLECT_RESULT_SUCCESS if succeeded, else return an error code.
*/
SpvReflectResult  spvReflectGetSpecializationInfo(VkSpecializationInfo& info);
```

shangjiaxuan:
Previously `BlockVariable`s size are statically calculated. Maybe should add an api for calculating based on specialization op result.
Also it seems adding the structure definition in header breaks build when including vulkan.h (duplicate definition). Should define opaque type in header.

shangjiaxuan:
According to spec `ARB_gl_spirv`, layout of buffers in should be static. Specialized array size should never exceed offset of next element. Thus, it's not necessary to recalculate offsets, only sizes, and mere size calculation do not need a new api.

chaoticbob:
This PR needs to be cleaned up. There are changes that are unrelated to SpecConstant support mixed in. It's difficult to see to which of these changes are relevant to the SpecConstant support and which are not.

There are some changes where `const` was added to parameter variables. While I'm not opposed to these, the lack of `const` wasn't an oversight. If you want to propose a change for this, it needs to come through on a separate PR. The same goes for the additional macros and other quality of life changes.

zbendefy:
Hi! I just wanted to comment that after integrating it into a custom project, I find it very useful. I'm just hoping for approval and a merge so I can build on this as part of the vulkan sdk. :)

chaoticbob:
In addition to the feedback I left:
- The giant macros need to become functions. The macros as they are cannot be debugged.
- The code format still needs work, it should follow the existing formatting with alignment and naming convention.
- Copy on returns should be avoided, the client code needs to make the allocation.
- There are few comments that are not clear on what exactly is happennig. 
- Comments should indicate a clear behavior even if that behavior is undefined.

chaoticbob (changes requested):

muraj:
Any update on this?  I would like to play with these in some of my shaders in order to dynamically change the work group size.  Thanks!

zbendefy:
Is there anything I can help out regarding this? :) I would really like to use spec constants via SPIRV-Reflect. I'm using shangjiaxuan's initial commit instead of the main branch, and its very useful.

spencer-lunarg:
@muraj @zbendefy  thanks for poking, I want to help move this forward, but something I am lacking are the use cases.

It would be helpful to have a "what is expected" vs "what it currently does" to better understand what part of spec constant support people are after

muraj:
Sorry for the delay.  Honestly, for my initial use case, just not barfing when specialization constants for the local group size are used would be fine.  When I first tried it, it failed to parse if any spec constant syntax was involved in the variable.  I had written up in full detail what happened somewhere.

Eventually I would like to retrieve more information about the variable definition for the local group size dimensions so I can consume any spirv given to modify the local group size and JIT at dispatch time to fit the local group size requested, caching more frequently used compilations for reuse later.

At the moment I'm just hard coding the specialization id, but I still need spirv-reflect to be able to parse the spirv in order to retrieve information about the entry point, etc.

spirv_reflect.c
```diff
@@ -3580,6 +3750,282 @@ static SpvReflectResult ParseExecutionModes(
...
+  }
+
+  p_module->specialization_constants = (SpvReflectSpecializationConstant*)calloc(p_module->specialization_constant_count, sizeof(SpvReflectSpecializationConstant));
+  if(IsNull(p_module->specialization_constants)){
+    return SPV_REFLECT_RESULT_ERROR_ALLOC_FAILED;
+  }
+
+  // deal with WorkGroupSize builtin
+  if (p_module->_internal->builtin_WorkGroupSize != INVALID_VALUE) {
+    SpvReflectPrvNode* p_node = FindNode(p_parser, p_module->_internal->builtin_WorkGroupSize);
```
mbechard, line 3864: What is this code block doing? Why is it overwriting the x value of the local size with the OpID that WorkGroupSize is held in?

spirv_reflect.c
```diff
@@ -3580,6 +3750,282 @@ static SpvReflectResult ParseExecutionModes(
...
+    for (uint32_t j = 0; j < p_module->entry_point_count; ++j) {
+      if (p_module->entry_points[j].spirv_execution_model == SpvExecutionModelKernel ||
```
mbechard, line 3866: What is this code block doing? Why is it overwriting the x value of the local size with the OpID that WorkGroupSize is held in?

spirv_reflect.c
```diff
@@ -3580,6 +3750,282 @@ static SpvReflectResult ParseExecutionModes(
...
+        p_module->entry_points[j].spirv_execution_model == SpvExecutionModelGLCompute) {
+        p_module->_internal->entry_point_mode_flags[j] = 4;
```
mbechard, line 3868: What is this code block doing? Why is it overwriting the x value of the local size with the OpID that WorkGroupSize is held in?

shangjiaxuan: When the code is written, specialization constants for workgroup size (aka `local_size_id_n`), as per spirv 1.5 specification, the compiler have two options for generating code. One is using `OpExecutionModeId`, and referencing resultid of three constant Ops. The other is using `WorkGroupSize` builtin, which takes a uvec3 argument resultid and overwrites the `OpExecutionMode` Op. When the code is written, `glslc` and other compilers are relying on the second (builtin) mode for generating code in glsl. The second mode used to be deprecated in spv1.5,  but later it seems the deprecation plan was moved to 1.6, if I remember correctly.

Reference:
https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html

Screenshot_20240229-074453

Screenshot_20240229-074419

Screenshot_20240229-074447

shangjiaxuan: For a reference of spv asm disassembled from glslc generated code that uses the builtin, see the sample test code included in pr:

https://github.com/shangjiaxuan/SPIRV-Reflect/blob/spec-constant/tests/spec_constant/test_orig.spv.dis

mbechard: Maybe the version of this PR I'm using is out of date, but in my case if a GLSL shader is using a spec constant, and is using the gl_WorkGroup variable, but the spec constant is **not** use to set the local_size, this code block still gets used and you end up with the incorrect value for local_size.

shangjiaxuan: > Maybe the version of this PR I'm using is out of date, but in my case if a GLSL shader is using a spec constant, and is using the gl_WorkGroup variable, but the spec constant is **not** use to set the local_size, this code block still gets used and you end up with the incorrect value for local_size.

Please give me sample glsl code and compiler options to compile it. Or you can give me the disassembled code. This can help clarify your problem and help me locate it.

shangjiaxuan: You should check out the entry mode flag that can be queried with `spvReflectGetEntryModeFlag`. In order to support specialized localsize, modes are added. The flag should have been in the localsize struct, but that would break abi and was requested to be moved to private member.

shangjiaxuan: The x member is overwritten to represent the uvec3 resultid if in builtin `WorkGroupSize` mode, and xyz are ovewritten as three resultid in `ExecutionModeId`.

shangjiaxuan: The rationale is that there might be specconstant arithmetic, whether scalar or vector. `WorkGroupSize` takes precedence over `ExecutionMode`, and the arguments used to be (1, 1, 1). It gives (16, 32, 1) default values with `WorkGroupSize` mainly for compatibility.

mbechard: hmm not sure I follow. In this case I want to get the values (16, 32, 1), since that is the actual local_size values. Where will those values reside when it's in this mode?

shangjiaxuan: You should use `spvReflectEvaluateResult` to evaluate the value of localsize.x resultid, for the current specialization constants in the evaluator (default values if you don't set them) to get the current uvec3.

mbechard: Ok, I can see how that'll result in me getter the correct answer. However, specialization constant's aren't used for the workgroup size in this example. The are hardcoded constants. It seems to me it should only enter into this mode if a specialization constant is actually used to set the workgroup size.

shangjiaxuan: It also baffles me why the compiler generates the builtin instruction. The older version compiler didn't do this if I remember correctly.

mbechard: I'm currently compiled again GLSLang 302a663a38, so it's not the most up to date version.
Another comment, I would think the reflection should put this result_id somewhere else, and try to resolve the correct local size in to the local_size variable. Re-using 'x' like this is confusing, and in general the user of reflection just wants the results, so asking them to use spvReflectEvaluateResult seems like an un-needed step.

A comment about the ABI (not directed at @shangjiaxuan), I don't understand why it would be requested that this tool which is only 2 files, should have a locked ABI. It's not completed enough to start hampering the interface with a locked ABI at this point.

RandomShaper:
> It would be helpful to have a "what is expected" vs "what it currently does" to better understand what part of spec constant support people are after

For what it's worth, I'll let you know why we need this in Godot.

- The API-agnostic layer of our rendering infrastructure use spec constant reflection for validation; namely, when it's asked to create a pipeline with some overrides for SC defaults, it validates the provided values are of the right type.
- Our Direct3D 12 renderer implementation still uses SPIR-V as an input for shaders. That means we have to emulate spec constants. We do that by patching DXIL resulting from transpilation. The matter is that knowing what the default values are allows us to optimize the patching process whenever a pipeline is created, by skipping what would otherwise be redundant operations.
- We're exposing shader reflection, including spec constants _with their default values_ to userland. That means user scripts (mostly tools, I guess) can use such data however they see fit.

PR #157, opened 2022-08-24
Fix spurious alloc error when there are no execution modes

RandomShaper:
`calloc()` returns `NULL` in some implementations when given a size of zero. This PR adds a guard for a case where that can happen in a valid shader to avoid a falsey `SPV_REFLECT_RESULT_ERROR_ALLOC_FAILED` result.


chaoticbob:
Great catch, thanks!

PR #162, opened 2022-12-05
Use same .clang-format as SPIRV-Tools

cassiebeckley:
This is just a suggestion, but SPIRV-Tools has a .clang-format and it might be nice to use the same formatting (and eventually maybe enforce it on PRs).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,6 @@
+---
+Language: Cpp
+BasedOnStyle: Google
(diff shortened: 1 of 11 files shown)


chaoticbob:
I merged your previous commit before looking at this one. I'm not sure if the conflict came as a result of the merge or if it was pre-existing (I don't think it was).

chaoticbob (approved):
LGTM

Issue #158, opened 2022-09-18
Adding additional word offsets to `SpvReflectBlockVariable`


alister-chowdhury:
Hey there,

It would be useful to be able to reflect the word offset of `offset` in UBOs, this allows you to reorder and compact structs.
This was a very trivial patch for me to implement locally, if it's something you wouldn't be against I can quickly put together a PR.

```diff
--- aspirv_reflect/spirv_reflect.c
+++ bspirv_reflect/spirv_reflect.c
@@ -2272,6 +2272,8 @@ static SpvReflectResult ParseDescriptorBlockVariable(
         ApplyArrayTraits(p_member_type, &p_member_var->array);
       }

+      p_member_var->word_offset.offset = p_type_node->member_decorations[member_index].offset.word_offset;
+
       p_member_var->type_description = p_member_type;
     }
   }

--- aspirv_reflect/spirv_reflect.h
+++ bspirv_reflect/spirv_reflect.h
@@ -378,6 +378,10 @@ typedef struct SpvReflectBlockVariable {
   uint32_t                          member_count;
   struct SpvReflectBlockVariable*   members;

+  struct {
+    uint32_t                          offset;
+  } word_offset;
+
   SpvReflectTypeDescription*        type_description;
 } SpvReflectBlockVariable;
```

chaoticbob:
Hi apologies for the late reply. Definitely not against it if it's useful to you - so definitely would love a PR. Just a couple of things:

1. Looking at `SpvReflectBlockVariable` - it seems like embedded struct isn't necessary, so a member called `word_offset` should be sufficient. I don't have a strong opinion on this.
2. Can you add it to the end of the struct?
3. 
Thanks!

PR #163, opened 2022-12-06
Adding word offset to SpvReflectBlockVariable

alister-chowdhury:
This change adds the word offset, of UBO offsets to the `SpvReflectBlockVariable` struct as part of #158.

I kept it in the `word_offset` struct, simply because you said you didn't have a strong opinion, but I can of course pop it out.


chaoticbob:
Thanks!

PR #171, opened 2023-04-18
Update spirv.h to the latest version in SPIRV-Headers.

dnovillo:
This adds support for the tile image extension.


dnovillo:
@chaoticbob This is failing the clang-format check because of the added `spirv.h` file.  Should we disable clang-format checks on this file?  I can re-format it, but it would create a significant diff against upstream.

dnovillo:
Fixed clang formatting issues.  @chaoticbob PTAL.  Thanks.

common/output_stream.cpp
```diff
@@ -279,6 +279,8 @@ std::string ToStringSpvDim(SpvDim dim) {
       return "Buffer";
     case SpvDimSubpassData:
       return "SubpassData";
+    case SpvDimTileImageDataEXT:
+      return "SpvDimTileImageDataEXT";
```
chaoticbob, line 283: I think generally the `Spv` prefix is dropped.

chaoticbob (changes requested):

common/output_stream.cpp
dnovillo: > I think generally the `Spv` prefix is dropped.

Thanks.  Fixed.

chaoticbob (approved):
LGTM

PR #174, opened 2023-04-27
add getter for the local_size

mbechard:


chaoticbob:
Hi, thanks for this addition. Would you mind adding a C version of the same function and having the C++ version calling the C version to keep the C and C++ APIs aligned?

mbechard:
The C API doesn't have any other parallel calls though, such as ones for `GetEntryPointShaderStage()`, `GetEntryPointName()` etc. It seems the C API just gets straight access to the `entry_points` structure via `spvReflectGetEntryPoint` and C users just access the struct directly. I was following the existing workflow, but honestly just giving access to the entry_point may be cleaner. For example I'm also needed access to output_vertices, which will require another getter.

mbechard:
Sorry, any confirmation on the feedback, do you want the parallel call added?

spirv_reflect.h
```diff
@@ -1479,6 +1479,10 @@ class ShaderModule {
   const char*                   GetEntryPointName(uint32_t index) const;
   SpvReflectShaderStageFlagBits GetEntryPointShaderStage(uint32_t index) const;
 
+  SpvReflectEntryPoint::LocalSize GetEntryPointLocalSize(uint32_t index) const {
```
chaoticbob, line 1482: It would be preferable here not to return a struct by copy - especially since it's an embeeded struct. 

Instead could it return `x`, `y`, `z` by pointer parameter and `index` is ranged check tgo make sure it's less than `SpvReflectShaderModule ::entry_point_count` and return `SPV_REFLECT_RESULT_ERROR_ELEMENT_NOT_FOUND` otherwise?

chaoticbob (changes requested):

chaoticbob:
Apologies, was out of pocket for a bit. I added some feedback.

mbechard:
Decided that using the C++ API is getting a bit cumbersome, so going to use the C one instead

PR #180, opened 2023-05-25
Add support for buffer pointers in buffers

greg-lunarg:
A buffer pointer is a pointer to a struct in the physical address space. The struct is decorated as is any buffer. These are currently generated by glsang for the GLSL buffer_reference feature.

This adds support for buffer pointers in buffers. Buffer pointers in buffers are treated as if the struct they point to is embedded in the buffer the pointer appears in at the point the pointer appears in. A new type flag bit is added to indicate a struct is actually a reference/pointer.

This commit does not add support for buffer pointers outside of buffers, for example, as local variables. That support is TBD.


greg-lunarg:
@chaoticbob I finally have this passing CI. Can you PTAL?

spirv_reflect.c
```diff
@@ -542,6 +551,52 @@ static SpvReflectTypeDescription* FindType(SpvReflectShaderModule* p_module, uin
   return p_type;
 }
 
+static SpvReflectPrvAccessChain* FindAccessChain(SpvReflectPrvParser* p_parser,
+                                                 uint32_t id) {
+  uint32_t ac_cnt = p_parser->access_chain_count;
+  for (uint32_t i = 0; i < ac_cnt; i++)
```
chaoticbob, line 557: Would you mind wrapping control flow statements in braces? Just to keep uniformity with the existing code? There's a few spots butI'll just leave thie comment on this one.

spirv_reflect.c
```diff
@@ -386,6 +390,11 @@ static SpvReflectResult ReadU32(
   return result;
 }
 
+#define UNCHECKED_READU32(parser, word_offset, value)             \
```
chaoticbob, line 393: Is this just for a a throw away read?

chaoticbob (changes requested):

spirv_reflect.c
greg-lunarg: Done.

spirv_reflect.c
greg-lunarg: No, not a throwaway read. It is a true read whose value is truly used. It is just does not have the check code so that it can be performed in other contexts besides the initial parsing context.

chaoticbob (approved):

chaoticbob:
Thanks!

spencer-lunarg:
I am just starting to look into why, but seems this PR regressed and now we segfault some shader, here is an example

https://github.com/SaschaWillems/Vulkan/blob/master/shaders/hlsl/conditionalrender/model.vert
https://github.com/SaschaWillems/Vulkan/blob/master/shaders/hlsl/conditionalrender/model.vert.spv

PR #182, opened 2023-05-26
Hold Component decoration value with Location

spencer-lunarg:
This fixes it so

```glsl
#version 450
layout(location = 0) in float a;
layout(location = 0, component = 2) in vec2 b;
void main(void) { }
```

can pass which component is being used, otherwise there is no way to tell there might be gaps in the components listed

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -130,6 +130,7 @@ typedef struct SpvReflectPrvDecorations {
   SpvReflectPrvNumberDecoration   binding;
   SpvReflectPrvNumberDecoration   input_attachment_index;
   SpvReflectPrvNumberDecoration   location;
+  SpvReflectPrvNumberDecoration   component;
   SpvReflectPrvNumberDecoration   offset;
   SpvReflectPrvNumberDecoration   uav_counter_buffer;
   SpvReflectPrvStringDecoration   semantic;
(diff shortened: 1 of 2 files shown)


spirv_reflect.c
```diff
@@ -2759,7 +2769,11 @@ static SpvReflectResult ParseInterfaceVariable(
   p_var->decoration_flags = ApplyDecorations(p_type_node_decorations);
   if (p_var_node_decorations != NULL) {
     p_var->decoration_flags |= ApplyDecorations(p_var_node_decorations);
+  } else {
```
chaoticbob, line 2772: nit: Would be preferable not to use the java style `if/else`.

chaoticbob (changes requested):
Just a nit.

This is new to men, wasn't aware you could specify a component in GLSL.

spirv_reflect.c
spencer-lunarg: sorry, what do you mean by "java style"? I just ran this against `clang-format` and this is what it gave

chaoticbob:
The windows image has been fixed on HEAD. Can you rebase this PR?

spirv_reflect.c
chaoticbob: Looks like the clang-format isn't as thorough. I'll update that.

chaoticbob (approved):

PR #190, opened 2023-06-12
Add enum for NV cooperative matrix type.

alelenv:
Add enum so as to not fail on SPV blob using OpTypeCooperativeMatrixNV

Have not added any tests since this is exactly similar to ray query type which can be only declared in Private/Function storage class and hence is never part of any reflection info


alelenv:
@chaoticbob : Please review and merge
Thanks!

chaoticbob (approved):

chaoticbob:
Thanks!

PR #196, opened 2023-06-21
Add define that disables the cxx implementation

cenomla:
Adds a define that disables the cxx implementation. This is useful for projects that use a cxx compiler but wish to only use the c api


chaoticbob (approved):

chaoticbob:
Thanks!

PR #199, opened 2023-07-01
Fix finding function parameter baseID of OpAccessChain

forenoonwatch:
This PR adds a fix to parsing push constant blocks whenever the code contains an `OpAccessChain` instruction with a baseID referencing an `OpFunctionParameter` instruction, as those were not handled at all.
For example, in this GLSL shader:
```glsl
#version 460

layout (push_constant) uniform PushConstants {
	float baz;
} constants;

// Keep the function from being marked unused
layout (location = 0) in vec2 valueA;
layout (location = 0) out float valueOut;

float foo(vec2 v) {
	return v.x;
}

void main() {
	valueOut = foo(valueA);
}
```


chaoticbob (approved):

chaoticbob:
Thanks!

Issue #177, opened 2023-05-24
Add runtime array info to SpvReflectBindingArrayTraits


spencer-lunarg:
From https://github.com/SaschaWillems/Vulkan/blob/master/shaders/hlsl/descriptorindexing/descriptorindexing.frag.spv

This returns an

```c++
SpvReflectBindingArrayTraits::dims_count = 1
SpvReflectBindingArrayTraits::dims = [ 0 ];
```

This shader uses a `OpTypeRuntimeArray` so it would be nice to know that instead of having to check for a zero `dim` value

spencer-lunarg:
I see this is noted in the header as 

>   // Each entry is: 0xFFFFFFFF for a specialization constant dimension,
  // 0 for a runtime array dimension, and the array length otherwise.
  
and it would be wasteful to hold another array to hold this info inside `SpvReflectArrayTraits`

PR #204, opened 2023-07-11
Use Enum instead of magic value for Array Dim

spencer-lunarg:
Was my issue in https://github.com/KhronosGroup/SPIRV-Reflect/issues/177 but having an Enum is nicer then just thinking there was a bug that it was set to `0`

### DIFF
--- a/spirv_reflect.h
+++ b/spirv_reflect.h
@@ -307,10 +307,17 @@ typedef struct SpvReflectImageTraits {
   SpvImageFormat                    image_format;
 } SpvReflectImageTraits;
 
+typedef enum SpvReflectArrayDimType {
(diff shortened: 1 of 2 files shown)


spirv_reflect.h
```diff
@@ -307,10 +307,17 @@ typedef struct SpvReflectImageTraits {
   SpvImageFormat                    image_format;
 } SpvReflectImageTraits;
 
+typedef enum SpvReflectArrayDimType {
+  SPV_REFLECT_ARRAY_DIM_RUNTIME = 0,                // OpTypeRuntimeArray
```
chaoticbob, line 311: nit: Assignment on enums should be aligned

chaoticbob (approved):
Looks good, just one nit.

chaoticbob (approved):

PR #205, opened 2023-07-13
Add way to access structs in array-of-structs

spencer-lunarg:
```
struct structX {
    int x1_;
    int x2_;
};
layout(location = 5) flat in structX int_b[2];
```

Before there was no was to get the `structX`'s `SpvReflectTypeDescription` without going through `_internal`

I would see it just skip by the struct actual type

```
type_description {
    op : OpTypeArray  
    type_name: 'structX'
    members: 
        type_description {
            op : OpTypeInt        
            struct_member_name: 'x1_'
        }
        type_description {
            op : OpTypeInt
            struct_member_name: 'x2_'
        }

}
```

This only happens with array-of-structs

keep `members` and `member_count` for backward compatibility sake, but these are now accessible going `struct_type_description->members`


chaoticbob (approved):

spirv_reflect.h
```diff
@@ -340,13 +341,19 @@ typedef struct SpvReflectTypeDescription {
     SpvReflectArrayTraits           array;
   } traits;
 
+  // If underlying type is a struct (ex. array of structs)
+  // this gives access to the OpTypeStruct
+  struct SpvReflectTypeDescription* struct_type_description;
+
+  // @deprecated use struct_type_description instead
   uint32_t                          member_count;
+  // @deprecated use struct_type_description instead
```
chaoticbob, line 350: How does struct_type_description account for the struct members? I don't see any code that parses the members vars.

chaoticbob (changes requested):

spirv_reflect.h
spencer-lunarg: So we call `ParseType` on all `OpTypeStructs` and build a `SpvReflectTypeDescription` for every struct already (and that is what is parsing the member vars)

Because SSA enforce ordering, when we get to the `ParseType` for a `OpTypeArray`, just calling `FindType()` will always find the struct the array is made up of

-----

printing from GDB

- `p reflection.m_module.input_variables[0].name` == `"int_b"`
- `p reflection.m_module.input_variables[0].type_description->type_name` == `"structX"`

so from the `SpvReflectTypeDescription` object

- `p type_description->struct_type_description.type_name` == `"structX"`
- `p type_description->struct_type_description.members[1].struct_member_name` = `"x2_"`
- `p type_description->members[1].struct_member_name` = `"x2_"`

chaoticbob (approved):
LGTM

PR #207, opened 2023-07-14
Add CI suite with SPIRV-Database

spencer-lunarg:
Help prevent regressions moving forward, just runs `spirv-reflect` on each shader (about 80,000 in total) and print the name (to know where it would crash)

The main goal of these is prevent crashing when using SPIRV-Reflect, not the output

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -89,6 +89,7 @@ int main(int argn, char** argv) {
   arg_parser.AddFlag("s", "stage", "");
   arg_parser.AddFlag("f", "file", "");
   arg_parser.AddFlag("fcb", "flatten_cbuffers", "");
+  arg_parser.AddFlag("ci", "ci", "");  // Not advertised
   if (!arg_parser.Parse(argn, argv, std::cerr)) {
     PrintUsage();
     return EXIT_FAILURE;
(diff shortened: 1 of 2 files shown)


chaoticbob (approved):

main.cpp
```diff
@@ -137,7 +139,10 @@ int main(int argn, char** argv) {
       return EXIT_FAILURE;
     }
 
-    if (print_entry_point || print_shader_stage || print_source_file) {
+    if (ci_mode) {
+      // CI Mode just is used to supress logs when running in CI
```
chaoticbob, line 143: nit: can you clarify this?

chaoticbob (changes requested):

main.cpp
spencer-lunarg:
```suggestion
      // When running CI we want to just test that SPIRV-Reflect doesn't crash,
      // The output is not important (as there is nothing to compare it too)
      // This hidden flag is here to allow a way to suppress the logging in CI to only the shader name
      // (otherwise the logs noise and GBs large)
```

How about this?

chaoticbob (approved):

chaoticbob (approved):

PR #208, opened 2023-07-14
Add default case to switch

s-perron:
When new source languages are added to spirv-headers, spirv-refelct will
not compile because not every possible case is handled by the switch.

I changed the `SpvSourceLanguageMax` case to be a default case. This
will make sure that spirv-refect can build with new version of
spirv-headers.

### DIFF
--- a/common/output_stream.cpp
+++ b/common/output_stream.cpp
@@ -110,7 +110,7 @@ std::string ToStringSpvSourceLanguage(SpvSourceLanguage lang) {
     case SpvSourceLanguageNZSL:
       return "NZSL";
 
-    case SpvSourceLanguageMax:
+    default:
       break;
   }
   // unhandled SpvSourceLanguage enum value
--- a/common/output_stream.cpp
+++ b/common/output_stream.cpp
@@ -89,8 +89,6 @@ std::string ToStringGenerator(SpvReflectGenerator generator) {
 
 std::string ToStringSpvSourceLanguage(SpvSourceLanguage lang) {
   switch (lang) {
-    case SpvSourceLanguageUnknown:
-      return "Unknown";
     case SpvSourceLanguageESSL:
       return "ESSL";
     case SpvSourceLanguageGLSL:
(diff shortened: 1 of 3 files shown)


spencer-lunarg (approved):
I was hitting this same warning, agree (but would like to return better values if this happens)

common/output_stream.cpp
```diff
@@ -110,7 +110,7 @@ std::string ToStringSpvSourceLanguage(SpvSourceLanguage lang) {
     case SpvSourceLanguageNZSL:
       return "NZSL";
 
-    case SpvSourceLanguageMax:
+    default:
       break;
   }
   // unhandled SpvSourceLanguage enum value
```
spencer-lunarg, line 116: instead of `return "???";` can we do something like `return "SpvSourceLanguage not found";`

spirv_reflect.c
```diff
@@ -5327,7 +5327,7 @@ const char* spvReflectSourceLanguage(SpvSourceLanguage source_lang)
     case SpvSourceLanguageSYCL           : return "SYCL";
     case SpvSourceLanguageHERO_C         : return "Hero C";
     case SpvSourceLanguageNZSL           : return "NZSL";
-    case SpvSourceLanguageMax:
+    default:
       break;
   }
   return "";
```
spencer-lunarg, line 5333:
```suggestion
  return "SpvSourceLanguage not found";
```

chaoticbob: I think the difficulty with this one is there exists a `SpvSourceLanguageUnknown` symbol - which currently returns the "Unknown".  My instinct is to say that the the case for that should be removed from the switch and and the default should break and "Unknown" should be returned. It looks like this is what `spirv-dis` does: https://github.com/KhronosGroup/SPIRV-Tools/blob/main/source/opcode.cpp#L64

chaoticbob (changes requested):
Consider "Unknown" similar to what `spirv-dis` does.

common/output_stream.cpp
s-perron: I return "Unknown" following @chaoticbob's suggestion in the other file.

spirv_reflect.c
s-perron: Done.

PR #210, opened 2023-07-19
Add LocalSizeId support

spencer-lunarg:
Before `LocalSizeId` would print out

`local size      : (0, 0, 0)`

If it finds a spec constant it looks like

`local size      : (Spec Constant, 16, 16)`

also I removed the default case returning an error as `VK_KHR_shader_subgroup_uniform_control_flow` was not parsing because it couldn't find the new `SubgroupUniformControlFlowKHR` execution mode, and I don't see the benefit here to try and track every new Execution Mode as we just are ignoring it anyway (and aren't `spirv-val`)


spirv_reflect.c
```diff
@@ -3513,32 +3531,7 @@ static SpvReflectResult ParseExecutionModes(
...
-        case SpvExecutionModeRoundingModeRTZ:
-        case SpvExecutionModeStencilRefReplacingEXT:
-        case SpvExecutionModeOutputLinesNV:
-        case SpvExecutionModeOutputPrimitivesNV:
-        case SpvExecutionModeOutputTrianglesNV:
-        case SpvExecutionModePixelInterlockOrderedEXT:
-        case SpvExecutionModePixelInterlockUnorderedEXT:
-        case SpvExecutionModeSampleInterlockOrderedEXT:
-        case SpvExecutionModeSampleInterlockUnorderedEXT:
+        default:
```
chaoticbob, line 3534: For this change, what happens if we encounter an execution mode that we haven't seen before? 

I see the addition of `SPV_REFLECT_EXECUTION_MODE_UNKNOWN` but not seeing it used.

spencer-lunarg: ya, can remove `SPV_REFLECT_EXECUTION_MODE_UNKNOWN` that was going to be for the `WorkgroupSize` support I decided to defer

> what happens if we encounter an execution mode that we haven't seen before?

Well before when we got something like `SpvExecutionModeOutputPoints` we just did nothing anyways, we track the values in the `spirv.h` so if they are using a valid execution mode it just works... if they aren't, well they somehow got pass all the other SPIR-V tools the would have blown up first

chaoticbob: > ya, can remove `SPV_REFLECT_EXECUTION_MODE_UNKNOWN` that was going to be for the `WorkgroupSize` support I decided to defer

Can leave it in if you plan to use it later. Not to fussed about it.

> > what happens if we encounter an execution mode that we haven't seen before?
> 
> Well before when we got something like `SpvExecutionModeOutputPoints` we just did nothing anyways, we track the values in the `spirv.h` so if they are using a valid execution mode it just works... if they aren't, well they somehow got pass all the other SPIR-V tools the would have blown up first

Got it. I guess we'd get a test would blow up or we'll get a bug report if it's too egregious.

chaoticbob (approved):

spirv_reflect.c
spencer-lunarg: > Can leave it in if you plan to use it later. Not to fussed about it.

Might be a while, better to just added when needed (also removed it)

Issue #212, opened 2023-07-20
Handle buffer reference with arrays to themselves


spencer-lunarg:
The following currently causes issues with how member type of structs are found

```glsl
#version 450
#extension GL_EXT_buffer_reference : enable

layout(buffer_reference) buffer T1;
layout(set = 3, binding = 1, buffer_reference) buffer T1 {
   layout(offset = 48) T1  c[2]; // stride = 8 for std430, 16 for std140
} x;

void main() {}
```

Can reproduce with test/glsl/buffer_handle_2.spv

chaoticbob:
It should look something like this:
```
          // size = 0, padded size = 0
          struct T1 {             
              ref T1  c[2];  // abs offset = 48, rel offset = 48, size = 16, padded size = 16, array stride = 8 UNUSED
          } x;
```

The logic would be someting like it's a buffer binding with a reference to something at offset=48. Size information for this might not be applicable *or* it's capped to the arary size. Need to understand what the shader developer's perspective on that would be.

Does `layout(buffer_reference)` itnroduce a new decoration or type? GLSL already causes a lot of confusion between the various OpTypeStruct for the block name, the block type, and the binding name. So not having sometype of anchoring information will make it even more annoying to parse.

cassiebeckley:
Looking at this closer, it seems that glslang compiles `T1` from the sample into `T1` and `T1_0`. If I manually update the SPIR-V to decorate these with the correct names, the output is

```
  Descriptor bindings: 1

    Binding 3.1
      spirv id : 5
      set      : 3
      binding  : 1
      type     : VK_DESCRIPTOR_TYPE_STORAGE_BUFFER (UAV)
      count    : 1
      accessed : false
      name     : x (T1)
          // size = 0, padded size = 0
          struct T1 {

              // abs offset = 48, rel offset = 48, size = 16, padded size = 16, array stride = 8 UNUSED
              ref struct T1_0 {

                  // abs offset = 0, rel offset = 0, size = 0, padded size = 0
                  ref struct T1_0 {
                  } T1_0[2];

              } c[2];

          } x;
```

which I _believe_ is correct (with the possible exception of `T1_0[2]` instead of `c_0[2]`.

spencer-lunarg:
[...]

PR #213, opened 2023-07-20
Fix regression in buffer reference crashing

spencer-lunarg:
The changes in https://github.com/KhronosGroup/SPIRV-Reflect/pull/180/ regressed the cases where a `GL_EXT_buffer_reference` points to itself

```glsl
layout(buffer_reference) buffer X {
    X a;
};
```

This PR prevents these cases from crashing from an infinite loop by tracking if we have seen the struct already when calling `ParseTypes`, `ParseDescriptorBlocks` and `ParsePushConstantBlocks`

I am still running into issues how array of recursive buffer reference are handled, but moved that to a separate issue to unblock the crashing
https://github.com/KhronosGroup/SPIRV-Reflect/issues/212


spirv_reflect.h
```diff
@@ -79,6 +79,7 @@ typedef enum SpvReflectResult {
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_BLOCK_MEMBER_REFERENCE,
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_ENTRY_POINT,
   SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_EXECUTION_MODE,
+  SPV_REFLECT_RESULT_ERROR_SPIRV_MAX_RECURSIVE,
```
chaoticbob, line 82: Can this be slightly more descriptive? Something like MAX_RECURSION_EXCEEDED?

spirv_reflect.c
```diff
@@ -1911,16 +1915,36 @@ static SpvReflectResult ParseType(
...
         // Parse type
         SpvReflectPrvNode* p_next_node = FindNode(p_parser, type_id);
-        if (IsNotNull(p_next_node)) {
-          result = ParseType(p_parser, p_next_node, NULL, p_module, p_type);
-        }
-        else {
+        if (IsNull(p_next_node)) {
           result = SPV_REFLECT_RESULT_ERROR_SPIRV_INVALID_ID_REFERENCE;
           SPV_REFLECT_ASSERT(false);
+        } else if (!found_recursion) {
```
chaoticbob, line 1946: nit: `else if` should be on next line

spirv_reflect.c
```diff
@@ -71,7 +71,8 @@ enum {
 
 // clang-format off
 enum {
-  MAX_NODE_NAME_LENGTH      = 1024,
+  MAX_NODE_NAME_LENGTH        = 1024,
+  MAX_RECURSIVE_POINTER_CHECK = 128
```
chaoticbob, line 75: nit: missing comma at end

Can you also add a one line comment on what this is used for?

chaoticbob (commented):
Not familiar buffer references, but new looks to stay in its lane so should be okay. Just a few small nits.

chaoticbob (approved):

PR #214, opened 2023-08-04
Fixed localSizeId always reading from x_node

ZXCqwe666:
SpvExecutionModeLocalSizeId case was writing to the entry point's localSize, 
by reading **only from the x_node into x, y, z fields**.

Its a slight typo that happened in [these commits](https://github.com/KhronosGroup/SPIRV-Reflect/pull/210/commits)


spencer-lunarg (approved):
what an aweful copy and paste error I made.. thanks for fixing this!

PR #220, opened 2023-09-04
Add geometry interface test

spencer-lunarg:
Adds a basic geometry test


tests/interface/geom_input_builtin_array.glsl
```diff
@@ -0,0 +1,21 @@
+#version 450 core
+
+in gl_PerVertex {
+    vec4 gl_Position;
+    float gl_CullDistance[3];
+} gl_in[];
```
spencer-lunarg, line 6: geometry shaders Input Builtin Block has an outer array

tests/interface/geom_input_builtin_array.spv.yaml
```diff
@@ -0,0 +1,208 @@
...
+    member_count: 0
+    members:
+  - &td1
+    id: 12
+    op: 28
+    type_name: 
+    struct_member_name: "gl_CullDistance"
+    storage_class: 0 # UniformConstant
+    type_flags: 0x20000008 # ARRAY FLOAT 
+    decoration_flags: 0x00000000 # NONE
```
spencer-lunarg, line 28: This **should** have a `built-in` label to it, but to be tracked as a future fix (if someone cares enough about geometry)

cassiebeckley (approved):

PR #223, opened 2023-09-10
Add Address Sanitization to CI build

spencer-lunarg:
- First commit just ran `dos2unix` on CMake file
- 2nd commit is adding an option to run with `-fsanitize=address` in CI to catch any potential issues (which currently there are none!)
  - Also run Release version in CI as that is what ends up in the SDK


spencer-lunarg:
CI with ASAN - 13 minutes
CI without ASAN - 3 minutes

So there is a noticeable increase in time, but feel with the frequency of PRs, this extra 10 minutes isn't going to block anyone

chaoticbob (approved):

Issue #222, opened 2023-09-08
Reflecting capabilities used by a shader more finely


mbechard:
If I'm given an arbitrary shader and I want to see if it runs on the local hardware, I can use reflection to see if any capabilities are being requested that the hardware doesn't support. However this isn't quite fine grained enough in many cases. For example if SpvCapabilityAtomicFloat32AddEXT is declared, I can see if `VK_EXT_shader_atomic_float `is supported as a first check. However the hardware may support atomic float operations on some data types and not others, as per the entries in `VkPhysicalDeviceShaderAtomicFloatFeaturesEXT`. For example on macOS `shaderImageFloat32AtomicAdd` is not currently supported.

Is there any path forward where we can see what kinds of operations are applied to data, such as flags on a reflected image descriptor saying it was used for an atomic float operation?

spencer-lunarg:
```
#version 450
#extension GL_EXT_shader_explicit_arithmetic_types_int64 : enable
#extension GL_EXT_shader_image_int64 : enable
layout(set = 0, binding = 0) buffer ssbo { uint64_t y; };
layout(set = 0, binding = 1, r64ui) uniform u64image2D z;
void main() {
    y = imageAtomicAdd(z, ivec2(1, 1), y);
}
```

will currently show

`accessed: 1`

but I can see how having a "type of access" flag to list if it was atomic could be helpful

Seems like something I could take a crack at

mbechard:
That would be amazing. A related request is this:
https://github.com/KhronosGroup/SPIRV-Reflect/issues/99
where it would be useful to know if the resource is used via a read and/or a write operation. This would allow more finely controlling the VkAccessFlags used on the resource before it's used.

spencer-lunarg:
@mbechard I have an implementation of this in https://github.com/KhronosGroup/SPIRV-Reflect/pull/224 (which I still need to invest more time before I feel comfortable merging it)

I would love feedback on it if you think it works as expected for your use case

mbechard:
Thanks! This looks great. Seems to work with a simple test case. Offhand I would expect the READ/WRITE flags to also be set when ATOMIC is set though. It allows for logic that is just looking if a resource is read or written to handle those flags regardless of if the  operation happens to be atomic or not

spencer-lunarg:
>  I would expect the READ/WRITE flags to also be set when ATOMIC is set though

[...]

PR #224, opened 2023-09-16
[WIP] Add Access Flags for DescriptorBinding

spencer-lunarg:
closes https://github.com/KhronosGroup/SPIRV-Reflect/issues/222

WIP because

1. Want to review and examine tests
2. Want to try to use the other `used_variables` allocation together with the new one added
3. Need to look into `OpCopy*` operations as currently don't think we walk it even for normal `accessed`

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -178,6 +178,22 @@ typedef struct SpvReflectPrvString {
 // clang-format on
 
 // clang-format off
+// There are a limit set of instructions that can touch an OpVariable,
+// these are represented here with how it was accessed
+// Examples:
+//    OpImageRead  -> OpLoad              -> OpVariable
+//    OpImageWrite -> OpLoad              -> OpVariable
+//    OpStore      -> OpAccessChain       -> OpVariable
+//    OpAtomicIAdd -> OpAccessChain       -> OpVariable
+//    OpAtomicLoad -> OpImageTexelPointer -> OpVariable
+typedef struct SpvReflectPrvAccessedVariable {
+  uint32_t               result_id;
+  uint32_t               variable_ptr;
+  SpvReflectAccessFlags  access_flags;
+} SpvReflectPrvAccessedVariable;
+// clang-format on
+
+// clang-format off
 typedef struct SpvReflectPrvFunction {
   uint32_t                        id;
   uint32_t                        callee_count;
@@ -185,6 +201,8 @@ typedef struct SpvReflectPrvFunction {
   struct SpvReflectPrvFunction**  callee_ptrs;
   uint32_t                        accessed_ptr_count;
   uint32_t*                       accessed_ptrs;
+  uint32_t                        accessed_variable_ptr_count;
+  SpvReflectPrvAccessedVariable*  accessed_variable_ptrs;
 } SpvReflectPrvFunction;
 // clang-format on
 
@@ -685,6 +703,7 @@ static void DestroyParser(SpvReflectPrvParser* p_parser)
       SafeFree(p_parser->functions[i].callees);
       SafeFree(p_parser->functions[i].callee_ptrs);
       SafeFree(p_parser->functions[i].accessed_ptrs);
+      SafeFree(p_parser->functions[i].accessed_variable_ptrs);
     }
 
     // Free access chains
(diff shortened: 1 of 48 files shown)


tests/access_chains/array_length_from_access_chain.spv.yaml
```diff
@@ -115,6 +115,7 @@ all_descriptor_bindings:
     block: *bv2 # 
     array: { dims_count: 0, dims: [] }
     accessed: 1
+    access_flags: 0x00000000 # NONE
```
spencer-lunarg, line 118: last thing left in PR is figure out why this is not consistent with `accessed`

spencer-lunarg: this is 

```
         %13 = OpAccessChain %_ptr_StorageBuffer__struct_3 %2
         %14 = OpArrayLength %uint %13 1
                     OpReturn
                     OpFunctionEn
```

and from looking at spec, this is not considered a `read`, just an access

tests/glsl/runtime_array_of_array_of_struct.spv.yaml
```diff
@@ -316,6 +317,7 @@ all_descriptor_bindings:
     block: *bv7 # ""
     array: { dims_count: 0, dims: [] }
     accessed: 1
+    access_flags: 0x00000000 # NONE
```
spencer-lunarg, line 320: seems 

```
OpStore -> OpAccessChain -> OpAccessChain -> OpVariable
```

is a valid thing... need to fix

spirv_reflect.h
```diff
@@ -262,6 +262,22 @@ typedef enum SpvReflectShaderStageFlagBits {
...
+NOTE: A variable may be "accessed" but still have SPV_REFLECT_ACCESS_NONE
+      Example is if there is a OpAccessChain, but then it is never used
+
+*/
+typedef enum SpvReflectAccessFlagBits {
+  SPV_REFLECT_ACCESS_NONE   = 0x00000000,
+  SPV_REFLECT_ACCESS_READ   = 0x00000001,
+  SPV_REFLECT_ACCESS_WRITE  = 0x00000002,
+  // Atomic will always also be marked as READ and WRITE
+  SPV_REFLECT_ACCESS_ATOMIC = 0x00000004,
```
chaoticbob, line 276: Is there ever a case where ATOMIC is used or checked by itself it it always inlcudes READ and WRITE? Wondering if there's a different way to express this.

spirv_reflect.c
```diff
@@ -1227,26 +1254,74 @@ static SpvReflectResult ParseFunction(
...
-      case SpvOpLoad:
+      } break;
       case SpvOpAccessChain:
       case SpvOpInBoundsAccessChain:
       case SpvOpPtrAccessChain:
       case SpvOpArrayLength:
       case SpvOpGenericPtrMemSemantics:
-      case SpvOpInBoundsPtrAccessChain:
+      case SpvOpInBoundsPtrAccessChain: {
+        CHECKED_READU32(p_parser, p_node->word_offset + 3,
```
chaoticbob, line 1264: Params should be all on one line or all on separate lines. Avoid paragraph style param packing. Same for the next few cases below.

spirv_reflect.c
```diff
@@ -1227,26 +1254,74 @@ static SpvReflectResult ParseFunction(
...
+                        p_func->accessed_ptrs[p_func->accessed_ptr_count]);
+
+        CHECKED_READU32(
+            p_parser, p_node->word_offset + 3,
+            p_func->accessed_variable_ptrs[p_func->accessed_ptr_count]
+                .variable_ptr);
```
chaoticbob, line 1270: For readability (as readable as C89 can be), can the member access just stay on the same line as their parent?

chaoticbob (changes requested):

spirv_reflect.h
spencer-lunarg: so I had it before as separate... https://github.com/KhronosGroup/SPIRV-Reflect/issues/222#issuecomment-1724754338 convinced me otherwise

spirv_reflect.c
spencer-lunarg: this was from clang-format

smonnnn:
Will this ever get merged? It'd be very useful.

s-perron (approved):
None of the reviews work on this project anymore. I'll approve since all of the issues raised earlier were reasonable style issue.

@spencer-lunarg You will need to rebase before you can merge.

PR #233, opened 2023-11-06
rerun clang-format with 132 line width

spencer-lunarg:
I plan after this lands to make a `.git-blame-ignore-revs` file with the commit, this way git blame on any modern git tool (and GitHub) will ignore this PR

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,4 +3,5 @@ Language: Cpp
 BasedOnStyle: Google
 DerivePointerAlignment: false
 SortIncludes: true
+ColumnLimit: 132
 ...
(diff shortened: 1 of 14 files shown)


cassiebeckley:
Is there a specific reason why you want to override the Google style guide default for this?

spencer-lunarg:
> Is there a specific reason why you want to override the Google style guide default for this?

Comments like [this](https://github.com/KhronosGroup/SPIRV-Reflect/pull/224#discussion_r1375642653) or while subjective, I feel this is not easy to read

image

and the style guide was always a "baseline" in my personal vision, we do this for the [Validation Layers too](https://github.com/KhronosGroup/Vulkan-ValidationLayers/blob/main/.clang-format)

also (subjective) I never understood why it was 80 char wide only to begin with

cassiebeckley (approved):
That's fair. I don't think we can override the style guide for internal projects at Google, but I have no personal objections to this :)

Issue #234, opened 2023-11-06
Feature request: reflect referenced members of a ByteAddressBuffer


danginsburg:
Given the following HLSL shader:

```
typedef uint Texture2DIndex_t;

struct MaterialData_t 
{
	float4 g_vTest;
	float2 g_vTest2;
	float3 g_vTest3;
	Texture2DIndex_t g_tTexture1;
	Texture2DIndex_t g_tTexture2;
	bool g_bTest1;
	bool g_bTest2;
};

static MaterialData_t _g_MaterialData;


ByteAddressBuffer g_MaterialData : register ( t4 , space1 )  ; 

struct Payload_t 
{ 
	float2 vTest;
	bool bTest;
	Texture2DIndex_t tTest;
	
}; 

struct PayloadShadow_t
{
	float m_flVisibility;
};

[ shader ( "closesthit" ) ] 
void ClosestHit0 ( inout Payload_t payload , in BuiltInTriangleIntersectionAttributes attrs ) 
{
	_g_MaterialData = g_MaterialData.Load<MaterialData_t>( InstanceIndex() );
 
    
    payload.vTest = _g_MaterialData.g_vTest2;
	payload.bTest = _g_MaterialData.g_bTest1;
	payload.tTest = _g_MaterialData.g_tTexture1;
	
    
} 

[ shader ( "anyhit" ) ] 
void AnyHit1 ( inout PayloadShadow_t payload , in BuiltInTriangleIntersectionAttributes attrs ) 
{
	_g_MaterialData = g_MaterialData.Load<MaterialData_t>( InstanceIndex() );
 
    { 
        payload . m_flVisibility = 0.0 ; 
        AcceptHitAndEndSearch ( ) ; 
    } 
} 
```

I would like to be able to use spirv-reflect to determine which members of g_MaterialData are statically referenced (the same as I would want to do for a cbuffer).  Currently that is not possible.

This shader can be lowered to spir-v using slang with:
```
slangc.exe byte_address_buffer_reflection.hlsl -o byte_address_buffer_reflection_slang.spv -profile lib_6_4 -entry ClosestHit0
```
Or with DXC:
```
dxc.exe -spirv -T lib_6_4 byte_address_buffer_reflection.hlsl -fspv-target-env=vulkan1.2 -Fo byte_address_buffer_reflection_dxc.spv
```

[...]

PR #236, opened 2023-11-14
Add ByteAddressBuffer support

spencer-lunarg:
prototype for https://github.com/KhronosGroup/SPIRV-Reflect/issues/234


spirv_reflect.h
```diff
@@ -440,11 +457,16 @@ typedef struct SpvReflectDescriptorBinding {
   SpvReflectBindingArrayTraits        array;
   uint32_t                            count;
   uint32_t                            accessed;
+  SpvReflectAccessFlags               access_flags;
   uint32_t                            uav_counter_id;
   struct SpvReflectDescriptorBinding* uav_counter_binding;
 
   SpvReflectTypeDescription*          type_description;
 
+  uint32_t                            is_byte_address_buffer;
```
chaoticbob, line 466: We should find some way to make this a flag so it can be used to identify other concrete types. ByteAddressBuffer is a concrete type on an SRV and there are a few others like it.

spencer-lunarg: sorry this is old (about to replace with new PR)

I created a 

```
typedef enum SpvReflectUserType {
  SPV_REFLECT_USER_TYPE_INVALID = 0,
  SPV_REFLECT_USER_TYPE_CBUFFER,
  SPV_REFLECT_USER_TYPE_TBUFFER,
  SPV_REFLECT_USER_TYPE_APPEND_STRUCTURED_BUFFER,
  ...
```

so this becomes an enum... basically will be a more general approach, this PR is **too** single focused to solve one task

spencer-lunarg:
closed for https://github.com/KhronosGroup/SPIRV-Reflect/pull/248

PR #241, opened 2023-12-25
Print Spec Constant value in type array by default

spencer-lunarg:
Instead of printing a mysterious `0xFFFFFFFF`, print the default spec constant value and make use of the `spec_constant_op_ids` field

so something like 

```
layout(constant_id = 3) const int SIZE = 2;
layout(set = 0, binding = 0, std430) buffer SSBO {
    float val[SIZE];
} ssbo;
```

now shows

```patch
// SpvReflectArrayTraits
dims_count = 0
- dims = [0xFFFFFFFF]
- spec_constant_op_ids = [constant_id]
+ dims = [2]
+ spec_constant_op_ids = [3]
```

(cc @muraj @zbendefy it would be awesome to get feedback if this helps solve your problems or not)

### DIFF
--- a/common/output_stream.cpp
+++ b/common/output_stream.cpp
@@ -1723,7 +1723,6 @@ void SpvReflectToYaml::WriteBlockVariable(std::ostream& os, const SpvReflectBloc
   os << t2 << "matrix: { ";
   os << "column_count: " << bv.numeric.matrix.column_count << ", ";
   os << "row_count: " << bv.numeric.matrix.row_count << ", ";
-  ;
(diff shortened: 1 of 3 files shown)


chaoticbob (approved):

zbendefy:
Hi!
Thanks this is definitely useful, I'm currently however missing the name of the constant in SpvReflectSpecializationConstant.

The way I set things up is that I use variable names, that later get resolved to a specialization constant ids, so it would help me a lot if I could also receive the name of the constant in 'SpvReflectSpecializationConstant'.

PR #242, opened 2023-12-25
Add baseline Spec Constant API support

spencer-lunarg:
the goal is not to dismiss https://github.com/KhronosGroup/SPIRV-Reflect/pull/197 or https://github.com/KhronosGroup/SPIRV-Reflect/pull/154 but instead create a more gradual first step to at least get "something" as a baseline for support

With this there is a `spvReflectEnumerateSpecializationConstants` which can be queried and returned a `SpvReflectSpecializationConstant` object.

For now `SpvReflectSpecializationConstant` is the bare minimum until it is better examined (and tested) what should be added, but then it will be a more simple additive process to what is already exposed

(cc @muraj @zbendefy it would be awesome to get feedback if this helps solve your problems or not)


chaoticbob (approved):

zbendefy:
Hi @spencer-lunarg !
Thanks this is definitely useful, I'm currently however missing the name of the constant in SpvReflectSpecializationConstant.

The way I set things up is that I use variable names, that later get resolved to a specialization constant ids, so it would help me a lot if I could also receive the name of the constant in 'SpvReflectSpecializationConstant'.

Issue #141, opened 2022-04-01
Storage image decorators are not parsed


juliusikkala:
I wanted to check whether a shader has marked a given storage image as readonly. I assume that this would result in `SPV_REFLECT_DECORATION_NON_WRITABLE` being visible somewhere, but this doesn't seem to be the case. Both `SpvReflectDescriptorBinding.block.decoration_flags` and `SpvReflectDescriptorBinding.type_description->decoration_flags` are zero regardless of the memory qualifier on the GLSL side.

I'm not particularly familiar with SPIR-V intricacies, so I don't really know which one of these two would be the right place for the flags for storage images. They're not really blocks, so maybe the type description? But it looks like the only `SpvReflectPrvNode`s with the correct decorator present are skipped on this line: https://github.com/KhronosGroup/SPIRV-Reflect/blob/5f6e13e06b8bd5a1eca190377b603bfd2aaa45ea/spirv_reflect.c#L1832

Maybe a `SpvReflectDecorationFlags decoration_flags;` could be added to `SpvReflectImageTraits` instead? Or to `SpvReflectDescriptorBinding` directly?

Also, looks like there are no tests for storage images. Here's a simple shader to reproduce the issue:
```glsl
#version 450
#pragma shader_stage(compute)

layout (local_size_x = 16, local_size_y = 16, local_size_z = 1) in;

layout(set = 0, binding = 0, rgba32f) uniform readonly image2D input_image;
layout(set = 0, binding = 1, rgba32f) uniform writeonly image2D output_image;

void main() {
    ivec2 p = ivec2(gl_GlobalInvocationID.xy);
    imageStore(output_image, p, imageLoad(input_image, p));
}
```
With the `spirv-reflect` program, there's no `NON_WRITABLE` in sight. I tested with the master branch in its current state.

chaoticbob:
Hi - added `decoration_flags` to `SpvReflectDescriptorBinding` per your suggestion. It seemed like the most logical place to go for `SPV_REFLECT_DECORATION_NON_WRITABLE` and `SPV_REFLECT_DECORATION_NON_READABLE`.

I haven't updated spirv-reflect to output NON_WRITABLE and NON_READABLE yet. Need to think of a way to output these without it looking too messy. Also affects the tests which is really a constraint on time right now.

Thanks for the example shader.  Let me know if you run into any issues.

PR #243, opened 2024-01-09
Promote member decorations to descriptor binding decorations

panos-lunarg:
This is basically a follow-up on #141.
Decorations like `readonly` or `writeonly` are being parsed but they are only applied to descriptor block variables.
This patch scans each descriptor's member decorations and those that are common to all members are promoted to the descriptor as well.


spirv_reflect.c
```diff
@@ -2039,6 +2039,34 @@ static SpvReflectResult ParseDescriptorBindings(SpvReflectPrvParser* p_parser, S
...
     p_descriptor->decoration_flags = ApplyDecorations(&p_node->decorations);
 
+    // Flags like non-writable and non-readable are found as member decorations only.
+    // If all members have one of those decorations set, promote the decoration up
+    // to the whole descriptor.
+    const SpvReflectPrvNode* p_type_node = FindNode(p_parser, p_type->id);
+    if (IsNotNull(p_type_node) && p_type_node->member_count) {
+      SpvReflectPrvDecorations common_flags = p_type_node->member_decorations[0];
+
+      for (uint32_t m = 1; m < p_type_node->member_count; ++m) {
```
chaoticbob, line 2049: Was curious why this starts at 1 intead of 0?

panos-lunarg: Member decoration at `0` is read in the line above. Then the rest are logical `&`ed one at a time. Result is if a decoration is `true` for all members, in the end it will remain `true`. These will be promoted to the whole descriptor. If there is only one decoration (at `0`) then its decorations will be promoted to the whole descriptor.

panos-lunarg:
With the intention of adding a test for this change I attempt to reuse the `non_writtable_image.spv` shader.
Because decorations were not being tested for descriptor bindings I had to touch all .yaml files. Hope it is the right way

spencer-lunarg (approved):
looks good to me, the tests (that aren't `NONE`) make sense to me looking at it

spencer-lunarg (changes requested):
After taking a second look at this, I am not sure if this is the correct move, if you have a shader like

```glsl
layout(set=0, binding=0) buffer storage_buffer {
    float a; // non-readable
    float b; // non-writable
} foo;
```

```swift

               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main"
               OpExecutionMode %main LocalSize 1 1 1
               OpMemberDecorate %storage_buffer 0 Offset 0
               OpMemberDecorate %storage_buffer 1 Offset 4
               OpMemberDecorate %storage_buffer 0 NonReadable
               OpMemberDecorate %storage_buffer 1 NonWritable
               OpDecorate %storage_buffer BufferBlock
               OpDecorate %foo DescriptorSet 0
               OpDecorate %foo Binding 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %float = OpTypeFloat 32
%storage_buffer = OpTypeStruct %float %float
%_ptr_Uniform_storage_buffer = OpTypePointer Uniform %storage_buffer
        %foo = OpVariable %_ptr_Uniform_storage_buffer Uniform
       %main = OpFunction %void None %3
          %5 = OpLabel
               OpReturn
               OpFunctionEnd
```

I don't think the desired result for spirv-reflect is to have the whole descriptor marked as both `NonReadable` and `NonWriteable`

panos-lunarg:
I assume you mean the following example:
```
layout(set=0, binding=0) buffer storage_buffer {
    writeonly float a; // non-readable
    readonly float b; // non-writable
} foo;
```

In the above example since none of the decorations are common between all member, none will be promoted to the whole descriptor.
In order to get _both_ `readonly` and `writeonly` promoted to the whole descriptor _all_ member must have the decorations. Something like this:
```
layout(set=0, binding=0) buffer storage_buffer {
    readonly writeonly float a; // both non-readable and non-writable
    readonly writeonly float b; // both non-readable and non-writable
} foo;
```
This indeed will mark `storage_buffer` as both `readonly` and `writeonly` but if glslang allows you to mark a member with both then I don't see the problem in promoting both to the descriptor

spencer-lunarg:
> In order to get both readonly and writeonly promoted to the whole descriptor all member must have the decorations.

Awww, I see, yes, you are doing `&=` on these

Could you quick add my example as a test case (show casing this code works like we intend it to) then I think I'm convinced this is ready to go

tests/glsl/readonly_writeonly.spv.yaml
```diff
@@ -0,0 +1,217 @@
...
+      - *bv2
+    type_description: *td3
+all_descriptor_bindings:
+  - &db0
+    spirv_id: 23
+    name: "foo"
+    binding: 0
+    input_attachment_index: 0
+    set: 0
+    decorations: 0x00000000 # NONE
```
spencer-lunarg, line 163: looks good, as expected!

common/output_stream.cpp
```diff
@@ -1808,6 +1809,8 @@ void SpvReflectToYaml::WriteDescriptorBinding(std::ostream& os, const SpvReflect
   //   uint32_t                            set;
   os << t1 << "set: " << db.set << std::endl;
   //   SpvReflectDescriptorType            descriptor_type;
+  os << t1 << "decorations: " << AsHexString(db.decoration_flags) << " # " << ToStringDecorationFlags(db.decoration_flags)
```
spencer-lunarg, line 1812: **FINAL** nit and happy to merge

this should be `"decoration_flags: "` to match the name of the header field

(idk why we still have these comments, but also adjust the `//   SpvReflectDescriptorType            descriptor_type;` above it)

spencer-lunarg (approved):

PR #244, opened 2024-01-10
TEST - Make sure CI is running correctly

spencer-lunarg:
inject failure and make sure CI fails


spencer-lunarg:
new flow works (fails) as intended

PR #248, opened 2024-01-12
Add ByteAddressBuffer support

spencer-lunarg:
attempt 2 of https://github.com/KhronosGroup/SPIRV-Reflect/pull/236 (closes https://github.com/KhronosGroup/SPIRV-Reflect/issues/234)

When `ByteAddressBuffer` are collapsed they start to look like

```swift
OpDecorateString %var UserTypeGOOGLE "byteaddressbuffer"
%var = OpVariable %22 StorageBuffer

%a = OpLoad %type %input
%b = OpIAdd %type %a %uint_16
%c = OpShiftRightLogical %type %b uint_2
%_ = OpAccessChain %e %var 0 %c
```

using the `UserType` we can detect which variable is used, from here we can build up an offset by working out the math

```swift
// offset + 16
%b = OpIAdd %type %a %uint_16
// offset >> 2
%c = OpShiftRightLogical %type %b uint_2
// offset == 4
```


spirv_reflect.c
```diff
@@ -3221,6 +3233,106 @@ static SpvReflectResult TraverseCallGraph(SpvReflectPrvParser* p_parser, SpvRefl
...
+  }
+
+  // It is costly to get the size before, so trade off a larger memory footprint (only if SPIR-V has ByteAddressBuffer user type)
+  if (IsNull(p_binding->byte_address_buffer_offsets)) {
+    p_binding->byte_address_buffer_offsets = (uint32_t*)calloc(SPV_REFLECT_MAX_BYTE_ADDRESS_BUFFER, sizeof(uint32_t));
+    if (IsNull(p_binding->byte_address_buffer_offsets)) {
+      return SPV_REFLECT_RESULT_ERROR_ALLOC_FAILED;
+    }
+  } else if (p_binding->byte_address_buffer_offset_count >= SPV_REFLECT_MAX_BYTE_ADDRESS_BUFFER) {
+    return SPV_REFLECT_RESULT_ERROR_ALLOC_FAILED;
```
danginsburg, line 3328: I am hitting this case with one of my shaders causing a downstream crash and byte address buffer reflection info not to work.  This needs to not be a hard limit or at least must larger.

spencer-lunarg: the allocation is now dynamic and should work now no matter how large your ByteAddressBuffer is

spirv_reflect.c
```diff
@@ -3221,6 +3233,106 @@ static SpvReflectResult TraverseCallGraph(SpvReflectPrvParser* p_parser, SpvRefl
...
+  // there is usually 2 (sometimes 3) instrucitons that make up the arithmetic logic to calculate the offset
+  SpvReflectPrvNode* arithmetic_node_stack[8];
+  uint32_t arithmetic_count = 0;
+
+  while (IsNotNull(p_next_node)) {
+    if (p_next_node->op == SpvOpLoad || p_next_node->op == SpvOpBitcast || p_next_node->op == SpvOpConstant) {
+      break;  // arithmetic starts here
+    }
+    arithmetic_node_stack[arithmetic_count++] = p_next_node;
+    if (arithmetic_count >= 8) {
```
danginsburg, line 3287: Should this check be above this?  Otherwise it will return if you have 8 operations when it could have succeeded (if you have exactly 8).  I also don't understand why there is such a low cap on this, it seems rather inflexible.  Would it be better to scan for how many first then allocate an array with a cap of like the size of the spir-v or something?

spencer-lunarg: so before it failed because I returned an error, not I return the same as "could not find" it it hits the cap

I guess there is no reason for a hard limit, it just I have yet to find a case where it is more then 2 here

> Would it be better to scan for how many first 

So the other issue is I do need "a cap" when doing this search anyways, there is no "correct" way for HLSL/Slang to flatten the `ByteAddressBuffer` so in the future, if some strange pattern arises, I still need some arbitrary number to stop at incase I get in some strange situation and don't want to crash

spirv_reflect.c
```diff
@@ -3296,30 +3409,57 @@ static SpvReflectResult ParseStaticallyUsedResources(SpvReflectPrvParser* p_pars
...
+        if (HasByteAddressBufferOffset(p_used_accesses[j].p_node, p_binding)) {
+          byte_address_buffer_offset_count++;
+        }
+      }
+    }
+
+    // only if SPIR-V has ByteAddressBuffer user type
+    if (byte_address_buffer_offset_count > 0) {
+      // possible not all allocated offset slots are used, but this will be a max per binding
+      p_binding->byte_address_buffer_offsets = (uint32_t*)calloc(byte_address_buffer_offset_count, sizeof(uint32_t));
```
danginsburg, line 3444: If there are more than one entrypoint, this reallocs without freeing the previous alloc, then there is a memory stomp in ParseByteAddressBuffer.

Something like this fixes it:

`	if (p_binding->byte_address_buffer_offset_count > 0) {
			uint32_t *prev_byte_address_buffer_offsets = p_binding->byte_address_buffer_offsets;
			p_binding->byte_address_buffer_offsets = (uint32_t *)calloc(byte_address_buffer_offset_count + p_binding->byte_address_buffer_offset_count, sizeof(uint32_t));
			memcpy(p_binding->byte_address_buffer_offsets, prev_byte_address_buffer_offsets, sizeof(uint32_t) * p_binding->byte_address_buffer_offset_count);
			SafeFree(prev_byte_address_buffer_offsets);
		}
		else {
			// possible not all allocated offset slots are used, but this will be a max per binding
			p_binding->byte_address_buffer_offsets = (uint32_t *)calloc(byte_address_buffer_offset_count, sizeof(uint32_t));
		}`

danginsburg:
I've tested the latest PR with our full shader corpus and can confirm it's working properly and producing proper ByteAddressBuffer offsets.  Looks good to merge to me.

chaoticbob (approved):
LGTM

PR #250, opened 2024-02-02
Parse name of specialization constants

zbendefy:
Parses names of specialization constants, also fixes parsing of spirv_id and constant_id (I think these were swapped when being parsed from the bytecode)

I still have to update the test yaml files to include the names, also I think this doesn't work yet for compute group size constants.

### DIFF
--- a/common/output_stream.cpp
+++ b/common/output_stream.cpp
@@ -2253,6 +2253,7 @@ void SpvReflectToYaml::Write(std::ostream& os) {
   os << t1 << "specialization_constants:" << std::endl;
   for (uint32_t i = 0; i < sm_.spec_constant_count; ++i) {
     os << t3 << "spirv_id: " << sm_.spec_constants[i].spirv_id << std::endl;
+    os << t3 << "name: " << sm_.spec_constants[i].name << std::endl;
     os << t3 << "constant_id: " << sm_.spec_constants[i].constant_id << std::endl;
   }
 
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -1652,8 +1652,13 @@ static SpvReflectResult ParseDecorations(SpvReflectPrvParser* p_parser, SpvRefle
       CHECKED_READU32(p_parser, p_node->word_offset + 2, decoration);
       if (decoration == SpvDecorationSpecId) {
         const uint32_t count = p_module->spec_constant_count;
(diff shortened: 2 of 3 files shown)


spirv_reflect.c
```diff
@@ -1652,8 +1652,13 @@ static SpvReflectResult ParseDecorations(SpvReflectPrvParser* p_parser, SpvRefle
       CHECKED_READU32(p_parser, p_node->word_offset + 2, decoration);
       if (decoration == SpvDecorationSpecId) {
         const uint32_t count = p_module->spec_constant_count;
-        CHECKED_READU32(p_parser, p_node->word_offset + 1, p_module->spec_constants[count].constant_id);
```
spencer-lunarg, line 1655: thanks for finding this bug!

common/output_stream.cpp
```diff
@@ -2253,6 +2253,7 @@ void SpvReflectToYaml::Write(std::ostream& os) {
   os << t1 << "specialization_constants:" << std::endl;
   for (uint32_t i = 0; i < sm_.spec_constant_count; ++i) {
     os << t3 << "spirv_id: " << sm_.spec_constants[i].spirv_id << std::endl;
+    os << t3 << "name: " << sm_.spec_constants[i].name << std::endl;
```
spencer-lunarg, line 2256: the lack of `SafeString` will end the printing

spirv_reflect.c
```diff
@@ -1652,8 +1652,13 @@ static SpvReflectResult ParseDecorations(SpvReflectPrvParser* p_parser, SpvRefle
...
-        CHECKED_READU32(p_parser, p_node->word_offset + 3, p_module->spec_constants[count].spirv_id);
+        CHECKED_READU32(p_parser, p_node->word_offset + 1, p_module->spec_constants[count].spirv_id);
+        CHECKED_READU32(p_parser, p_node->word_offset + 3, p_module->spec_constants[count].constant_id);
+        SpvReflectPrvNode* target_node = FindNode(p_parser, p_module->spec_constants[count].spirv_id);
+        if (IsNull(target_node)) {
```
spencer-lunarg, line 1658: it is valid for this to be null

spencer-lunarg (commented):
@zbendefy Thanks for making this PR, I took a look and there was just a edge case with `LocalSizeId`

https://github.com/KhronosGroup/SPIRV-Reflect/pull/251 has your commit with the exception added on that if you want to take that instead

PR #256, opened 2024-03-27
Fix SpvReflectEntryPoint::output_vertices being stomped by other execution modes

handsomematt:
These execution modes set and stomp output_vertices with invalid data since they have no extra operands.

```
[4]  OpExecutionMode %2 OutputVertices 128
[5]  OpExecutionMode %2 Invocations 1
[6]  OpExecutionMode %2 InputPoints
```

With this SPIRV generated by DXC InputPoints would stomp the OutputVertices value to a non-existent value


spencer-lunarg (approved):
Thanks, this is clearly a bug, not sure how these `Execution Mode` got here

Issue #257, opened 2024-04-04
Error parsing compute shader SPIR-V generated by `glslang` 14.1.0.


corporateshark:
Error parsing compute SPIR-V generated by `glslang` 14.1.0.

![image](https://github.com/KhronosGroup/SPIRV-Reflect/assets/2510143/a1ab3c2c-88b8-4a60-af44-ed600d93ac7e)

`base_node->op == SpvOpBitcast`

SPIR-V compute shader: [dump_2.zip](https://github.com/KhronosGroup/SPIRV-Reflect/files/14864339/dump_2.zip)
SPIRV-Reflect: [vulkan-sdk-1.3.280.0](https://github.com/KhronosGroup/SPIRV-Reflect/releases/tag/vulkan-sdk-1.3.280.0)

spencer-lunarg:
can reproduce with

```glsl
#version 460
#extension GL_EXT_buffer_reference2 : require
#extension GL_EXT_buffer_reference_uvec2 : require

layout(buffer_reference) buffer VertexBuffer;
layout(buffer_reference, buffer_reference_align = 16, std430) buffer VertexBuffer {
    int x;
};

layout(push_constant, std430) uniform PerFrameData {
    uvec2 bufferId;
} pc;

void main() {
    VertexBuffer(pc.bufferId).x = 0;
}
```

the core issue is we aren't handling cases of using the `OpBitcast` introduced by `GL_EXT_buffer_reference_uvec2` everywhere correctly

corporateshark:
@spencer-lunarg what should we do about it?

spencer-lunarg:
sorry not fully back to my normal work setup yet... just took a second look and now double guessing myself what is the "right" answer is (will have more time tomorrow to take a look at this and get a proper fix with testing)

for now, a "temp fix" would be to go

```patch
diff --git a/spirv_reflect.c b/spirv_reflect.c
index 92f8b33..cb6e178 100644
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -555,6 +555,9 @@ static uint32_t FindBaseId(SpvReflectPrvParser* p_parser, SpvReflectPrvAccessCha
       case SpvOpFunctionParameter: {
         UNCHECKED_READU32(p_parser, base_node->word_offset + 2, base_id);
       } break;
+      case SpvOpBitcast: {
+        return 0;
+      } break;
       default: {
         assert(false);
       } break;
```

[...]

PR #259, opened 2024-04-07
Fix asserting from uvec2 buffer reference bitcast

spencer-lunarg:
closes https://github.com/KhronosGroup/SPIRV-Reflect/issues/257

confirmed doing more research that we are safe to return `0` for a push constant that tries to use `uvec2` to pass in a buffer reference handle


spirv_reflect.c
```diff
@@ -529,17 +529,19 @@ static SpvReflectTypeDescription* FindType(SpvReflectShaderModule* p_module, uin
...
     }
   }
   return 0;
 }
 
-static uint32_t FindBaseId(SpvReflectPrvParser* p_parser, SpvReflectPrvAccessChain* ac) {
-  uint32_t base_id = ac->base_id;
+// Access Chains mostly have their Base ID pointed directly to a OpVariable, but sometimes
+// it will be through a load and this funciton handles the edge cases how to find that
+static uint32_t FindAccessChainBaseVariable(SpvReflectPrvParser* p_parser, SpvReflectPrvAccessChain* p_access_chain) {
```
spencer-lunarg, line 543: while here, updated this function name and `ac` to `p_access_chain` for consistency

corporateshark (approved):

corporateshark:
> confirmed doing more research that we are safe to return 0 for a push constant that tries to use uvec2 to pass in a buffer reference handle

Based on this logic, will we obtain the correct byte size for push constants?

spencer-lunarg:
>  will we obtain the correct byte size for push constants?

Yes, as shown from the shader in the issue, the `bufferId` is still properly being accounted for (and properly marked as `used`)

```
          // size = 128, padded size = 128
          struct PerFrameData {               
              column_major float4x4 MVP;       // abs offset =   0, rel offset =   0, size = 64, padded size = 64 UNUSED
              uint2                 bufferId;  // abs offset =  64, rel offset =  64, size =  8, padded size =  8
              uint                  textureId; // abs offset =  72, rel offset =  72, size =  4, padded size =  4 UNUSED
              float                 time;      // abs offset =  76, rel offset =  76, size =  4, padded size =  4
              uint                  numU;      // abs offset =  80, rel offset =  80, size =  4, padded size =  4
              uint                  numV;      // abs offset =  84, rel offset =  84, size =  4, padded size =  4
              float                 minU;      // abs offset =  88, rel offset =  88, size =  4, padded size =  4 UNUSED
              float                 maxU;      // abs offset =  92, rel offset =  92, size =  4, padded size =  4 UNUSED
              float                 minV;      // abs offset =  96, rel offset =  96, size =  4, padded size =  4 UNUSED
              float                 maxV;      // abs offset = 100, rel offset = 100, size =  4, padded size =  4 UNUSED
              uint                  P1;        // abs offset = 104, rel offset = 104, size =  4, padded size =  4
              uint                  P2;        // abs offset = 108, rel offset = 108, size =  4, padded size =  4
              uint                  Q1;        // abs offset = 112, rel offset = 112, size =  4, padded size =  4
              uint                  Q2;        // abs offset = 116, rel offset = 116, size =  4, padded size =  4
              float                 morph;     // abs offset = 120, rel offset = 120, size =  4, padded size =  8
          } pc;  
```

-----

The issue going on here was because inside a buffer reference struct, you can have things pointing to itself (ex, linked list)

```glsl
layout(buffer_reference) buffer VertexBuffer {
    VertexBuffer next_pointer;
};
```

and we have logic to handle this recursion, but a Push Constants block, it can't be a `buffer_reference` block by definition, so there is no way to have it point to itself, so we don't need to chase down the buffer reference chain and can just return early

PR #267, opened 2024-06-26
Correctly report liveness of variables passed as function parameters

### LINKED ISSUE #102: Bindings missing from reflection data if passed by parameter
When I compile this shader and run it through spirv_reflect, I get reflection data as I expect

```
#version 450

layout (set = 0, binding = 1) uniform sampler smp;
layout (set = 1, binding = 3) uniform texture2D tex;

layout (location = 0) out vec4 out_color;

vec4 normal_map(
    texture2D t,
    sampler s,
    vec2 uv
) {
    // *** NOT using the parameters, just accessing tex/smp directly ***
    return texture(sampler2D(tex, smp), uv);
}

void main() {
    out_color = normal_map(tex, smp, vec2(1.0, 1.0));
}
```

If I use this code instead, the bindings are not in the reflection data.

```
#version 450

layout (set = 0, binding = 1) uniform sampler smp;
layout (set = 1, binding = 3) uniform texture2D tex;

layout (location = 0) out vec4 out_color;

vec4 normal_map(
    texture2D t,
    sampler s,
    vec2 uv
) {
    // *** Using parameters t and s instead of "global" tex/smp ***
    return texture(sampler2D(t, s), uv);
}

void main() {
    out_color = normal_map(tex, smp, vec2(1.0, 1.0));
}
```

I'm going through rust bindings, but this is the code I'm calling:
```rust
    let enumerated_entry_points = shader_module.enumerate_entry_points()?;
    for entry_point in enumerated_entry_points {
        for descriptor_set in &entry_point.descriptor_sets {
            for binding in &descriptor_set.bindings {
                println!("binding {} {} {}", binding.name, binding.set, binding.binding);
            }
        }
    }
```

Is this expected behavior? I expected that these shaders would produce identical reflection data.

daniel-story:
Closes #102


spirv_reflect.c
```diff
@@ -3439,6 +3472,44 @@ static SpvReflectResult ParseByteAddressBuffer(SpvReflectPrvParser* p_parser, Sp
   return SPV_REFLECT_RESULT_SUCCESS;
 }
 
+static SpvReflectResult ParseFunctionParameterAccess(SpvReflectPrvParser* p_parser, uint32_t callee_function_id,
+                                                     uint32_t function_parameter_index, uint32_t* accessed) {
```
chaoticbob, line 3476: Function params and vars that are pointers should have `p_` prefix so `accessed` should be `p_accessed`.

spirv_reflect.c
```diff
@@ -1152,8 +1170,25 @@ static SpvReflectResult ParseFunction(SpvReflectPrvParser* p_parser, SpvReflectP
...
+      case SpvOpFunctionParameter: {
+        CHECKED_READU32(p_parser, p_node->word_offset + 2, p_func->parameters[p_func->parameter_count]);
+        (++p_func->parameter_count);
+      } break;
       case SpvOpFunctionCall: {
         CHECKED_READU32(p_parser, p_node->word_offset + 3, p_func->callees[p_func->callee_count]);
+        const uint32_t result_index = p_node->word_offset + 2;
+        for (uint32_t j = 0, parameter_count = p_node->word_count - 4; j < parameter_count; j++) {
+          const uint32_t ptr_index = p_node->word_offset + 4 + j;
+          SpvReflectPrvAccessedVariable* access_ptr = &p_func->accessed_variables[p_func->accessed_variable_count];
```
chaoticbob, line 1182: Does `access_ptr` indicate that it's accessing a pointer or is the `_ptr` suffix to indicate that this is variable is a pointer? If it's the latter, then it should be `p_access` instead.

chaoticbob (changes requested):
Thanks for the PR, looks good overall. Just a two nits.

spirv_reflect.c
daniel-story: This line was copied and pasted from the SpvOpLoad..SpvOpImageTexelPointer case below (see line 1204). I can change it on this line, but perhaps it's better to fix both lines at once in a separate PR?

spirv_reflect.c
daniel-story: This has been fixed.

spirv_reflect.c
chaoticbob: Ahh, apologies I didn't see that. You're right, that should be handled in a separate PR, not critical.

chaoticbob (approved):

chaoticbob (approved):

tests/user_type/byte_address_buffer_0.spv.yaml
```diff
@@ -101,7 +101,7 @@ all_descriptor_bindings:
     accessed: 1
     uav_counter_id: 4294967295
     uav_counter_binding:
-    ByteAddressBuffer offsets: [4, 5, 11, 13]
+    ByteAddressBuffer offsets: [13, 5, 11, 4]
```
spencer-lunarg, line 104: little concerned why this was flipped here, but not in `user_type/byte_address_buffer_3.spv.yaml`

spencer-lunarg: less concern because the order wasn't important/required

spencer-lunarg (approved):

PR #268, opened 2024-06-30
Fix specialization_constants items

dpwiz:
`specialization_constants` block is missing item markers and the resulting YAML is treated like a map, failing on duplicate keys (`spirv_id`).


spencer-lunarg (approved):

spencer-lunarg:
@dpwiz sorry - I just realized this and https://github.com/KhronosGroup/SPIRV-Reflect/pull/274 were doing the same thing and I merged that one in first

Issue #271, opened 2024-07-03
segmentation fault when parsing shader containing buffer-references


fabian-lunarg:
using spirv-reflect from latest vulkan-sdk-1.3.283.0 (or HEAD), ubuntu 22.04, x86_64:

```
spirv-reflect g_buffer_vert.spv
[1]    171736 segmentation fault (core dumped)  spirv-reflect
```

I'm almost sure this a regression, but could not bisect when it got introduced.
the shader uses an array of buffer-references, which I believe is the trigger for the segfaults.
other shaders containing the same pattern also segfault.

```
layout(buffer_reference, scalar) readonly buffer VertexBufferPtr { packed_vertex_t v[]; };
layout(binding = BINDING_VERTICES, set = 0, scalar) readonly buffer Vertices { VertexBufferPtr vertex_buffers[]; };
```

I'm able to hotfix with a simple nullptr-check, but unsure if there is more to be done (@spencer-lunarg ?).
PR coming up

[g_buffer_vert.zip](https://github.com/user-attachments/files/16084383/g_buffer_vert.zip)

PR #272, opened 2024-07-03
Add nullptr-check to prevent segfault.

fabian-lunarg:
- Occurred in shaders containing arrays of buffer-references
- closes https://github.com/KhronosGroup/SPIRV-Reflect/issues/271


fabian-lunarg:
this hotfix became obsolete, instead use https://github.com/KhronosGroup/SPIRV-Reflect/pull/273

PR #273, opened 2024-07-08
Fix runtime array of BufferDeviceAddresses

spencer-lunarg:
closes https://github.com/KhronosGroup/SPIRV-Reflect/issues/271

@fabian-lunarg the fix was even more stupid simple then you might have imagined... added test case and your shader in issue works now as well


dnovillo (approved):
Thanks!

PR #274, opened 2024-07-08
fix yaml generation for specialization constants

geor-kasapidi:


spencer-lunarg:
you will need to run `python3 tests/build_golden_yaml.py` to update the tests otherwise CI is just going to fail if I run it

geor-kasapidi:
@spencer-lunarg done

spencer-lunarg (approved):

PR #276, opened 2024-07-25
Add support for shared library build

makkarpov:
This PR adds support for building SPIRV-Reflect as a shared library


spencer-lunarg (approved):
LGTM - @chaoticbob mentioned how we should add a blurb to the `README.md` to mention this is possible and which cmake options to set

makkarpov:
Ok, I will add mentions in README.md that shared library build is possible.

Also, to be on the safe side, I will also perform DLL testing on Windows with actual application. It should work, but its better to check.

PR #277, opened 2024-08-04
Update CMake to 3.16

corporateshark:
Fixed this CMake 3.25 issue:
```
CMake Deprecation Warning at SPIRV-Reflect/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of CMake.
  ```


spencer-lunarg (approved):
to my knowledge, cmake `3.17.2` is the version the rest of the Vulkan tools use, so 3.16 should be good

PR #282, opened 2024-10-04
Fix generating invalid YAML files

Dunedan:
YAML 1.0 demands that the [separator between a YAML directive and its value is a colon](https://yaml.org/spec/1.0/#id2562252). This wasn't the case for the YAML files generated by spirv-reflect, as it used a space as separator instead. This resulted in tools parsing YAML stricly not being able to process the YAML files generated by spirv-reflect.

Newer YAML versions demand a [space as separator between YAML directive and its value](https://yaml.org/spec/1.1/#id895217) instead.

To fix this, this commit increases the YAML version to 1.1. This is done instead of replacing the separator to improve compatibility with tooling, as many tools don't support YAML 1.0 and require YAML 1.1 as minimal version (see the list of tools on https://yaml.org/).

### DIFF
--- a/common/output_stream.cpp
+++ b/common/output_stream.cpp
@@ -2148,7 +2148,7 @@ void SpvReflectToYaml::Write(std::ostream& os) {
   const std::string t2 = Indent(indent_level + 2);
   const std::string t3 = Indent(indent_level + 3);
 
-  os << "%YAML 1.0" << std::endl;
+  os << "%YAML 1.1" << std::endl;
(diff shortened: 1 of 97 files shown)


spencer-lunarg (approved):
@chaoticbob this seems fine with me, are you aware of any automated tools that were consuming these YAML that **would** break from this

spencer-lunarg:
been a week, going to merge this since in theory it was already broken pretending to be 1.0 doing 1.1 things

PR #286, opened 2024-11-11
Fix undefined enums

s-perron:
There are a few places where an enum value is purpoposly getting a value
that is not one of the enum values. This is causing a problem when
testing. I have a quick fix to make those values `int` instead of the
enum type.

The variables could contain `-1` as a special value to indicate that no
storage class or builtin is used. We cannot change the enum class to add
`-1` because it is defined in spirv headers.

### DIFF
--- a/spirv_reflect.h
+++ b/spirv_reflect.h
@@ -432,7 +432,7 @@ typedef struct SpvReflectInterfaceVariable {
   const char*                         semantic;
   SpvReflectDecorationFlags           decoration_flags;
   
-  // The builtin id if the variable is a builtin, and -1 otherwise.
+  // The builtin id (SpvBuiltIn) if the variable is a builtin, and -1 otherwise.
   int                                 built_in;
   SpvReflectNumericTraits             numeric;
   SpvReflectArrayTraits               array;
--- a/spirv_reflect.h
+++ b/spirv_reflect.h
@@ -393,7 +393,7 @@ typedef struct SpvReflectTypeDescription {
   const char*                       type_name;
   // Non-NULL if type is member of a struct
   const char*                       struct_member_name;
-  SpvStorageClass                   storage_class;
+  int                               storage_class;
   SpvReflectTypeFlags               type_flags;
   SpvReflectDecorationFlags         decoration_flags;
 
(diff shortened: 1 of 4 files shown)


spirv_reflect.h
```diff
@@ -393,7 +393,7 @@ typedef struct SpvReflectTypeDescription {
   const char*                       type_name;
   // Non-NULL if type is member of a struct
   const char*                       struct_member_name;
-  SpvStorageClass                   storage_class;
+  int                               storage_class;
```
spencer-lunarg, line 396: I would rather try and keep the interface the same... maybe the answer is to use `StorageClassMax` and typedef it as 

`SpvStorageClass invalid_storage_class = StorageClassMax`

where we are purposely setting it to be invalid

spencer-lunarg: I guess nvm, this would be wrong because it would show everything

spencer-lunarg: maybe what we need is 

```suggestion
// If an invalid Storage Class will be set to -1 because SpvStorageClass has no "invalid" value
  int                               storage_class;
```

s-perron: Yup, I don't think using `Max` is a good idea because its value will potentially change everytime the spirv headers are updated. I don't know if that create a problem for tooling using spirv-reflect.

The other possibility is to change the header file to spirv.hpp11, which uses enum classes. I believe that would make it okay to use `-1`. However, I don't know how much we want to change.

spencer-lunarg: thinking more, I think the best we can do is have this as an `int` but leave a comment above each line (as we might introduce compiler warning/errors with this PR) that `storage_class` and `built_in` can be casted to `SpvStorageClass/SpvBuiltIn` if the `int` value is not `-1`

s-perron: The enum class would not work, some of the files are written in C.

spirv_reflect.h
```diff
@@ -429,7 +431,9 @@ typedef struct SpvReflectInterfaceVariable {
   SpvStorageClass                     storage_class;
   const char*                         semantic;
   SpvReflectDecorationFlags           decoration_flags;
-  SpvBuiltIn                          built_in;
+  
+  // The builtin id if the variable is a builtin, and -1 otherwise.
```
spencer-lunarg, line 435:
```suggestion
  // The builtin id (SpvBuiltIn) if the variable is a builtin, and -1 otherwise.
```

cassiebeckley (approved):

spencer-lunarg (approved):

Issue #290, opened 2024-12-04
Push constant size incorrectly padded to 16 alignment


danginsburg:
Given the following shader (https://shader-slang.com/slang-playground/?target=SPIRV&code=eJwrLikqTS5RcEksSazmUlBQyMwrUaiw5qq15uKKji7LtrIqKC3OiE_OzysuScwriY0Fq1QACTpDxay5gsJDUitKSotSDV1sgPrtFNLj_UtLCkpLQIbkleaWZBSlJqYUaxjqAKFmLFdZfmaKQnJ-LlBFqm9iZp6GJhfIbpiuaINYBVsUK_RALgIAaEQ55Q):

```
struct Data{
   int x;
};

[[vk::push_constant]] Data pushConstant;
RWTexture1D<int> g_Output;

[numthreads(1,1,1)]
void computeMain()
{
  g_Output[0] = pushConstant.x;
}
```

With spir-v output:
```
; SPIR-V
; Version: 1.5
; Generator: Khronos; 40
; Bound: 23
; Schema: 0
               OpCapability Image1D
               OpCapability Shader
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %computeMain "main" %pushConstant %g_Output
               OpExecutionMode %computeMain LocalSize 1 1 1
               OpSource Slang 1
               OpName %Data_std430 "Data_std430"
               OpMemberName %Data_std430 0 "x"
               OpName %pushConstant "pushConstant"
               OpName %g_Output "g_Output"
               OpName %computeMain "computeMain"
               OpDecorate %Data_std430 Block
               OpMemberDecorate %Data_std430 0 Offset 0
               OpDecorate %g_Output Binding 0
               OpDecorate %g_Output DescriptorSet 0
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
        %int = OpTypeInt 32 1
%Data_std430 = OpTypeStruct %int
%_ptr_PushConstant_Data_std430 = OpTypePointer PushConstant %Data_std430
      %int_0 = OpConstant %int 0
%_ptr_PushConstant_int = OpTypePointer PushConstant %int
         %13 = OpTypeImage %int 1D 2 0 0 2 R32i
%_ptr_UniformConstant_13 = OpTypePointer UniformConstant %13
      %v4int = OpTypeVector %int 4
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
[...]

PR #291, opened 2024-12-04
Fixes issue #290 - set is_parent_rta for ParseDescriptorBlockVariableSizes

danginsburg:
so that they don't get padded to 16-byte size.

Changes the output for the shader in the bug report as follows:
```
generator       : Khronos SPIR-V Tools Assembler
source lang     : Unknown
source lang ver : 1
source file     :
entry point     : main (stage=CS)
local size      : (1, 1, 1)


  Push constant blocks: 1

    0:
      spirv id : 2
      name     : pushConstant (Data_std430)
          // size = 4, padded size = 4
          struct Data_std430 {
              int x; // abs offset = 0, rel offset = 0, size = 4, padded size = 4
          } pushConstant;



  Descriptor bindings: 1

    Binding 0.0
      spirv id : 3
      set      : 0
      binding  : 0
      type     : VK_DESCRIPTOR_TYPE_STORAGE_IMAGE (UAV)
      count    : 1
      accessed : true
      name     : g_Output
```

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -4016,7 +4016,7 @@ static SpvReflectResult ParsePushConstantBlocks(SpvReflectPrvParser* p_parser, S
     }
 
     p_push_constant->name = p_node->name;
-    result = ParseDescriptorBlockVariableSizes(p_parser, p_module, true, false, false, p_push_constant);
+    result = ParseDescriptorBlockVariableSizes(p_parser, p_module, true, false, true, p_push_constant);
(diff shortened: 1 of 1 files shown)


spencer-lunarg:
This will need to update the golden test YAML for other things

```
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release
cmake --build build
python tests\build_golden_yaml.py
```

The case you had this works, but I am still looking into if the BufferDeviceAddress handles make sense or not for the new size

PR #303, opened 2025-01-30
Don't freeze up when no args are passed in

spencer-lunarg:
someone pointed out if you go `./spirv-reflect` it just hangs, this just has this print the help


main.cpp
```diff
@@ -90,7 +90,7 @@ int main(int argn, char** argv) {
   arg_parser.AddFlag("f", "file", "");
   arg_parser.AddFlag("fcb", "flatten_cbuffers", "");
   arg_parser.AddFlag("ci", "ci", "");  // Not advertised
-  if (!arg_parser.Parse(argn, argv, std::cerr)) {
+  if (!arg_parser.Parse(argn, argv, std::cerr) || arg_parser.GetArgs().empty()) {
```
spencer-lunarg, line 93: I just realized this doesn't account for `stdin` actually

PR #315, opened 2025-07-03
Added more SPIR-V source languages

Luna5ama:
See https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_source_language


spencer-lunarg:
this looks good, the CI might be broken for other reasons, will take a look and merge in if I find it an unrelated issue (will also get around to review your other 2 PRs today, thanks for the contributions!)

PR #316, opened 2025-07-03
Added support for multiple entrypoints for CLI yaml print

### LINKED ISSUE #313: Command Line Tool doesn't support multiple entrypoints
#56 added such support to the library, but the command line tool only shows one entrypoint.

Luna5ama:
Fixed #313


tests/user_type/byte_address_buffer_0.spv.yaml
```diff
@@ -103,7 +103,7 @@ all_descriptor_bindings:
     accessed: 1
     uav_counter_id: 4294967295
     uav_counter_binding:
-    ByteAddressBuffer offsets: [13, 5, 11, 4]
+    ByteAddressBuffer offsets: [4, 5, 11, 13]
```
Luna5ama, line 106: For whatever reason, the ordering is messed up on Windows with MSVC. It was supposed to be sorted. But that would be another issue and PR.
https://github.com/KhronosGroup/SPIRV-Reflect/blob/main/spirv_reflect.c#L3669-L3674

PR #322, opened 2025-09-11
Enforcing stricter compilation warnings and fixing enum typing

M2-TE:
On Linux (gcc, clang) and MacOS (clang), the flags `-Wextra` and `-Wpedantic` can be used in addition to the not-so-aptly-named `-Wall` to catch additional mishaps.

Clang caught one minor issue with the enums `INVALID_VALUE` and `SPV_REFLECT_EXECUTION_MODE_SPEC_CONSTANT`, which are constants set to `0xFFFFFFFF`. Stricly speaking, the hexadecimal is an `unsigned int`, representing a value that the enum (int) cannot represent. To solve this, I simply cast the hexadecimal to an int, which creates an int with all bits set. Enums with an underlying type other than `int` are a specific extension in C, so this would be the more "correct" solution.

GCC 15 caught an issue regarding the cast from `INVALID_VALUE` to `SpvStorageClass`, writing to `storage_class` of `SpvReflectTypeDescription`.

https://github.com/KhronosGroup/SPIRV-Reflect/blob/94e1c5ea4ffe4a4c951b322c6217cb822876de14/spirv_reflect.h#L390-L406

 `storage_class` is already an int, so the cast is both unnecessary and strictly speaking an overflow.

Some comparisons with the enums have them simply cast to `unsigned int` beforehand to resolve issues with sign differences (most flags seem to be of `unsigned int`).

### DIFF
--- a/spirv_reflect.c
+++ b/spirv_reflect.c
@@ -1779,7 +1779,7 @@ static SpvReflectResult ParseType(SpvReflectPrvParser* p_parser, SpvReflectPrvNo
         SpvReflectTypeDescription* p_member_type = &(p_type->members[i]);
         p_member_type->id = (uint32_t)INVALID_VALUE;
         p_member_type->op = (SpvOp)INVALID_VALUE;
-        p_member_type->storage_class = (SpvStorageClass)INVALID_VALUE;
+        p_member_type->storage_class = INVALID_VALUE;
       }
     } else {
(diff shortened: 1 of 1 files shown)


E-Parker (approved):
Seems perfectly fine. I was about to implement exactly this fix but thought to check the pull requests first.

PR #344, opened 2026-05-20
Add SPV_EXT_descriptor_heap reflection support

chaoticbob:
This PR adds support for SPV_EXT_descriptor_heap and updated support for SPV_KHR_untyped_pointers. Test cases are a best effort since full compiler support for SPV_EXT_descriptor_heap is still in flight. A follow-up PR is planned once compiler support lands. Decorations, builtins, and instructions are parsed for compliance. These may need to be adjusted in the future.


spencer-lunarg (approved):
looks good enough to me, just need to solve the CI issue

PR #350, opened 2026-08-03
cmake: add shared library option

wolfv:
Add an opt-in `SPIRV_REFLECT_SHARED_LIB` CMake option alongside the existing static-library option.

The shared target installs as `libspirv-reflect`, installs the public header, and uses CMake automatic symbol exports on Windows. Existing defaults and the static target remain unchanged.

Tested by configuring, building, and installing the shared-only target locally. This is also exercised across Linux, macOS, and Windows by conda-forge: https://github.com/conda-forge/spirv-reflect-feedstock

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(spirv-reflect)
 OPTION(SPIRV_REFLECT_EXECUTABLE     "Build spirv-reflect executable" ON)
 
 OPTION(SPIRV_REFLECT_STATIC_LIB     "Build a SPIRV-Reflect static library" OFF)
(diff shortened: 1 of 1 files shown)


wolfv:
recheck

spencer-lunarg (approved):

spencer-lunarg:
@wolfv this looks good... not sure why github is not happy about the CLA, I tried to "recheck" it and still seems to not be working... can you try signing it again (github has REALLY had issues with the CLA server recently across the various repos I work on)

wolfv:
OK, now CLA is green!

PR #355, opened 2026-09-13
Make the GCC deprecation-message check actually test the GCC version

Chang-Jin-Lee:
Fixes #336.

`GCC_VERSION` is not predefined by GCC, so `#if GCC_VERSION >= 40500` has always compared 0 against 40500 and taken the branch below it. Every deprecated symbol in the public header loses its message on GCC:

```
warning: ‘spvReflectGetShaderModule’ is deprecated [-Wdeprecated-declarations]
```

and with this change it reads

```
warning: ‘spvReflectGetShaderModule’ is deprecated: renamed to spvReflectCreateShaderModule [-Wdeprecated-declarations]
```

`-Wundef` names the cause on its own: `"GCC_VERSION" is not defined, evaluates to 0`, at `spirv_reflect.h:49`. Clang and MSVC were never affected, since they are handled by the branches above.

The issue suggests defining `GCC_VERSION`. I spelled the comparison out instead — this is a public header, and `GCC_VERSION` is a common enough spelling in other projects that defining it here could collide with a consumer's own. The `>= 40500` gate is left alone rather than dropped, so the intent of the original code is unchanged.

Checked on GCC 13.3.0: the message appears in C and in C++ under `-Wall -Wextra`, the `-Wundef` warning is gone, and `test-spirv-reflect` is 3678 passing. The edit sits inside the file's `clang-format off` region, so the formatting job has nothing to reflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


spencer-lunarg (approved):

spencer-lunarg:
@Chang-Jin-Lee can you make sure the CLA is signed for this repo so I can merge this

(if you did, try https://cla-assistant.io/ to retrigger it)

Chang-Jin-Lee:
Sorry for the slow reply. The CLA is mine to clear and it is still unsigned — `license/cla` is the only thing not green here, everything else passed. I am going through the signing flow now and will retrigger it as you suggested once it is done.

